PR #25, opened 2015-05-29
new client func should return type client, not marathon

nicgrayson:


nicgrayson:
@gambol99 I see you merge my readme PR. Any chance you can get this one merged too? If you need an example or have questions let me know.

sherzberg:
Or does Client just not implement all of the Marathon methods?

gambol99:
Admittedly the use of the Marathon interface isn't really required, I could have happily stuck to using the Client struct. Still, it's just habit to do interface / implementation (good old java days).

Issue #23, opened 2015-05-27
Listing apps with labels


bobrik:
Marathon can filter apps with specific labels: mesosphere/marathon#1416. You can also embed tasks into apps to avoid extra requests with `embed=apps.tasks`.

I propose add an argument to `Marathon.Applications()`, like this:

```go
type ApplicationsOptions struct {
	Embed string
	Cmd   string
	Label string
}
```

I can come up with a PR.

bobrik:
It probably worth thinking how to extend that to other api endpoints that take params.

gambol99:
Have to admit, I'm kinda inclined to provide a generic Params like struct, wrapping a map[string]interface and allow programmer to make the choice adding or not adding options. That way we don't have to add, remove, amend options per endpoint as Marathon evolves and changes it's API; The only danger I foresee down that road would be if the option alters the assumed response (i.e. struct) your expecting to decode into??

bobrik:
Well, `map[string]interface{}` is probably too much, `url.URL.Query()` is `map[string][]string`. I think we can do just `map[string]string` or use some interface that can convert itself to `map[string]string`.

Take a look at [elastic](https://github.com/olivere/elastic/blob/78b87db2a9c53fb29acfaca1c16764d66a7ac519/query.go). I can try to implement something like `Options` and iterate on it in PR. What do you think?

gambol99:
Sounds good to me

gambol99:
closing ... as the changes we're merged

PR #26, opened 2015-05-29
WIP: Add query string params for API

bobrik:
Related to #23.

Can we reuse `url.Values` directly, @gambol99? Should we add it to every api call?

### DIFF
--- a/application.go
+++ b/application.go
@@ -19,6 +19,7 @@ package marathon
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


gambol99:
sorry, just saw this .... Yeah, using url.Values make's sense to me ... rather than reinventing the same thing under different name ..

bobrik:
Take a look now.

Few notes:

* `KillApplicationTasks` passes params in body, probably can use `url.Values` instead
* `DeleteDeployment` has unused `force` param, `url.Values` can be used as well

bobrik:
@gambol99 mind to take a look? I rebased agains latest master.

PR #30, opened 2015-06-08
Remove unnecessary else blocks/indentations; Fix a few grammar errors…

Atheatos:
Some changes I made while reading the code. Unless I'm mistaken, the else blocks weren't necessary and removing them might make the code nicer to read due to fewer indentations.


Atheatos:
Looks like I was mostly mistaken. Some of them should be okay, though.

PR #40, opened 2015-06-26
Add a ListTasks function for getting a list of task ids

Atheatos:
Essentially a copy and paste of ListApplications()


gambol99:
the build failure is unrelated ... I'm guessing they have changed the url.Parse() method in tip .. merging in now

PR #44, opened 2015-08-13
Added support for marathon event app_terminated_event

hamo:
This event does not appear in the event doc, but appear in the code. src/main/scala/mesosphere/marathon/event/Events.scala

It will happen when one app terminated.

[Review on Reviewable](https://reviewable.io/reviews/gambol99/go-marathon/44)


gambol99:
perfect ... merging now

PR #55, opened 2015-09-20
Added option to force scale operation

Raffo:
With this PR I added the option to force the scale operation. This basically means adding "force=true" in the query string. 

As a side note, I also removed a duplicate entry of myself from the AUTHORS file :-)

### DIFF
--- a/application.go
+++ b/application.go
@@ -465,12 +465,18 @@ func (client *Client) RestartApplication(name string, force bool) (*DeploymentID
 // Change the number of instance an application is running
 // 		name: 		the id used to identify the application
 // 		instances:	the number of instances you wish to change to
-func (client *Client) ScaleApplicationInstances(name string, instances int) (*DeploymentID, error) {
+//    force: used to force the scale operation in case of blocked deployment
+func (client *Client) ScaleApplicationInstances(name string, instances int, force bool) (*DeploymentID, error) {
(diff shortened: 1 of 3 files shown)


application.go
```diff
@@ -465,12 +465,18 @@ func (client *Client) RestartApplication(name string, force bool) (*DeploymentID
...
 // 		name: 		the id used to identify the application
 // 		instances:	the number of instances you wish to change to
-func (client *Client) ScaleApplicationInstances(name string, instances int) (*DeploymentID, error) {
+//    force: used to force the scale operation in case of blocked deployment
+func (client *Client) ScaleApplicationInstances(name string, instances int, force bool) (*DeploymentID, error) {
 	client.log("ScaleApplicationInstances(): application: %s, instance: %d", name, instances)
 	changes := new(Application)
 	changes.ID = validateID(name)
 	changes.Instances = instances
-	uri := fmt.Sprintf("%s/%s", MARATHON_API_APPS, trimRootPath(name))
```
gambol99: .. we could probably use the below instead of if/else
```go
uri := fmt.Sprintf("%s/%s?force=%t", MARATHON_API_APPS, trimRootPath(name), force)
```

Raffo: You're totally right, I don't know why I did it like this... updated the PR. :+1:

PR #59, opened 2015-10-09
VersionInfo values on app objects are now parsed.

xavi-:
Marathon v0.11.0 added a property called `versionInfo` to app objects.  This PR makes it accessible.  We'd like to use `LastConfigChangeAt` for an internal project.


burnsed:
LGTM

PR #63, opened 2015-10-10
code cleanup

gambol99:
- removing alot of the if/else from the code, making the code easier to read
- adding the TasksHealthy and TasksUnhealthy to Application v0.11


application.go
```diff
@@ -304,10 +314,12 @@ func (client *Client) SetApplicationVersion(name string, version *ApplicationVer
 	client.log("SetApplicationVersion() setting the application: %s to version: %s", name, version)
 	uri := fmt.Sprintf("%s/%s", MARATHON_API_APPS, trimRootPath(name))
 	deploymentId := new(DeploymentID)
-	if err := client.apiPut(uri, version, deploymentId); err != nil {
+	err := client.apiPut(uri, version, deploymentId)
+	if err != nil {
```
eicca: AFAIK the `if err := doSomething; err != nil { ..` is acceptable pattern in golang: http://blog.golang.org/error-handling-and-go
In this post all the single error assignments (`err :=`) are written in the `if err :=` manner.

eicca:
The code looks much better now :+1:

application.go
gambol99: Agreed .... for single error returns it makes sense to keep the if err:= doSomething; err != nil { ... i'll revert that line

PR #65, opened 2015-10-11
Fix/formatting

gambol99:


application.go
```diff
@@ -21,22 +21,22 @@ import (
 	"fmt"
 	"net/url"
 	"time"
+	"github.com/golang/glog"
 )
 
 var (
+	// ErrApplicationExists ... the application already exists in the marathon
 	ErrApplicationExists = errors.New("The application already exists in marathon, you must update")
```
eicca: I think this error is not used anywhere.

application.go
```diff
@@ -72,132 +72,136 @@ type Application struct {
...
 	Version string `json:"version"`
 }
 
+// VersionInfo ... the application versioning details from marathon
 type VersionInfo struct {
 	LastScalingAt      string `json:"lastScalingAt,omitempty"`
 	LastConfigChangeAt string `json:"lastConfigChangeAt,omitempty"`
 }
 
+// NewDockerApplication ... create a default docker application
```
eicca: I think the comments should be in the form `X does stuff`. When `golint` says that their should be in the form `X ...` it means that instead of `...` there should be some actual text.
For example: https://github.com/golang/glog/blob/master/glog.go#L164
In this case it could be: `NewDockerApplication creates a default docker application`.
It's a minor thing, however it might be a bit easier to read.

application.go
```diff
@@ -72,132 +72,136 @@ type Application struct {
...
 	application.Container = NewDockerContainer()
 	return application
 }
 
-// The name of the application i.e. the identifier for this application
-func (application *Application) Name(id string) *Application {
-	application.ID = validateID(id)
-	return application
+// Name ... set the name of the application i.e. the identifier for this application
+func (r *Application) Name(id string) *Application {
```
eicca: Again just a minor thing: usually in this case the first letter is used for self reference. You can look at the glog example as well: https://github.com/golang/glog/blob/master/glog.go#L222
In this case it could be `func (a *Application) Name(id string)`.

application.go
```diff
@@ -21,22 +21,22 @@ import (
 	"fmt"
 	"net/url"
 	"time"
+	"github.com/golang/glog"
```
eicca: It would be nice to have more formatted logging. And I think most of the companies have their own logging standard. In this it would be really awesome to use the interface (as I see only `Infof` is needed) and provide glog as default implementation.
In this case everybody can pass his own implementation or just use default glog.
What do you think about it?

eicca:
Since the project is `golint` compatible now, what do you think about adding `golint` (or `go vet`) to the travis?
I checked some popular OSS projects and found that `go vet` is usually included:
- https://github.com/hashicorp/consul/blob/master/Makefile
- https://github.com/kubernetes/kubernetes/blob/master/hack/vet-go.sh

when `golint` is not. I don't know what's the reason since the code must be linted anyway.

client.go
```diff
@@ -28,196 +28,182 @@ import (
...
 	ErrInvalidArgument = errors.New("The argument passed is invalid")
-	/* error return by marathon */
-	ErrMarathonError = errors.New("Marathon error")
-	/* the operation has timed out */
+	// ErrTimeoutError ... the operation has timed out
 	ErrTimeoutError = errors.New("The operation has timed out")
 )
 
-type Client struct {
+type marathonClient struct {
```
eicca: Thanks, having an exported interface implementation was a bit confusing before :)

client.go
```diff
@@ -226,7 +212,7 @@ func (client *Client) marshallJSON(data interface{}) (string, error) {
 	return string(response), err
 }
 
-func (client *Client) unMarshallDataToJson(stream io.Reader, result interface{}) error {
+func (r *marathonClient) unMarshallDataToJSON(stream io.Reader, result interface{}) error {
```
eicca: I think it should be `unmarshalDataToJSON` (one `l`, one word) according to this https://golang.org/pkg/encoding/json/#Unmarshal

client.go
```diff
@@ -28,196 +28,182 @@ import (
...
+// Ping ... Pings the current marathon endpoint (note, this is not a ICMP ping, but a rest api call)
+func (r *marathonClient) Ping() (bool, error) {
+	if err := r.apiGet(MARATHON_API_PING, nil, nil); err != nil {
 		return false, err
 	}
 	return true, nil
 }
 
-func (client *Client) marshallJSON(data interface{}) (string, error) {
+func (r *marathonClient) marshallJSON(data interface{}) (string, error) {
```
eicca: `marshalJSON`: https://golang.org/pkg/encoding/json/#Marshal

cluster_test.go
```diff
@@ -23,45 +23,39 @@ import (
...
-func GetFakeCluster() {
-	if cluster == nil {
-		cluster, _ = NewMarathonCluster(FAKE_MARATHON_URL)
-	}
-}
-
 func TestUrl(t *testing.T) {
-	GetFakeCluster()
-	assert.Equal(t, cluster.Url(), FAKE_MARATHON_URL)
+	cluster, _ := newCluster(FAKE_MARATHON_URL)
```
eicca: It might make sense to rename `FAKE_MARATHON_URL` to `fakeMarathonURL` for consistency.

const.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package marathon
 
 const (
+	DEBUG_LEVEL        = 10
```
eicca: `DebugLevel`?
As I know constants follow the same rule as usual variables: https://github.com/golang/glog/blob/master/glog.go#L100

application.go
gambol99: First off, cheers for the feedback btw ... it is very much welcomed. Yep, i'll removed a error as it not used

application.go
gambol99: fixing now :-)

application.go
gambol99: I might do this is another PR - I guess is this is more preference that anything else ... sticking with the same letter every time is less confusing r = reciever etc .. None the less, to stick with convention i'll move over to it in another PR

cluster_test.go
gambol99: good point ... changing to remove the exports

eicca:
Thanks for making the changes :+1:

PR #68, opened 2015-10-16
UpdateApplication returns DeploymentID.

eicca:
According to https://mesosphere.github.io/marathon/docs/rest-api.html#put-v2-apps-appid marathon returns deployment information after updating an app.

Additionally, duplicate `put` endpoint was removed from `methods.yml`.


gambol99:
:+1: ... merging now ..

PR #69, opened 2015-10-17
Return result after waitOnRunning.

eicca:
Now if the option `waitOnRunning` is passed the method will return the result of API call instead of `nil`.

As a side note: I would rather drop `waitOnRunning` option, so these methods have only single responsibility. User can call then `WaitOnApplication` or `ApplicationOK` by himself.


gambol99:
You have a point on waitOnRunning. i'll chunk into in a #issue

PR #71, opened 2015-10-22
Remove some omitempty

rubbish:
I ran into a few of these when I was trying to update an marathon app's attributes to empty. With the `omitempty`, if they're empty, they do not get sent. So removing that fixed it. Also, I removed the `omitempty` on the `MaxConsecutiveFailures` as 0 is a valid option (marathon then doesn't stop the task).


gambol99:
LGTM ... merging

PR #72, opened 2015-10-22
Deployments confusion

rubbish:
This clears up `Deployments` and `DeploymentIDs` on the application. `Deployments` now matches what the JSON says and `app.DeploymentIDs()` is a nice wrapper around that.

This one builds on #71 as otherwise they'll conflict.

### DIFF
--- a/application.go
+++ b/application.go
@@ -55,7 +55,7 @@ type Application struct {
 	BackoffSeconds        float64             `json:"backoffSeconds,omitempty"`
 	BackoffFactor         float64             `json:"backoffFactor,omitempty"`
 	MaxLaunchDelaySeconds float64             `json:"maxLaunchDelaySeconds,omitempty"`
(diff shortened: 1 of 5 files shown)


eicca:
Should the interface be changed as well: https://github.com/Banno/go-marathon/blob/deployments-confusion/client.go#L107?

application.go
```diff
@@ -191,13 +191,13 @@ func (r *Application) HasHealthChecks() bool {
 }
 
 // Deployments retrieves the application deployments ID
```
eicca: The comment won't pass `golint` now. Can you change it please to `DeploymentIDs does something`?

rubbish:
Good call. I'll get those changed up.

rubbish:
@eicca I don't believe the client or marathon interface has to change. This is only on the Application interface.

eicca:
@rubbish I'm sorry, I got confused by the methods with the same name :)

PR #73, opened 2015-10-26
Adds customisable CallbackUrl

elliot:
Solves issue where the library is run within a Docker container.

When inside of a container, it will attempt to (correct) bind to the eth0 interface on the docker0 bridge (172.16.0.0/16). This will result in a non routable callback url being sent to Marathon. This change gives user the ability to specify a custom endpoint, failing back to the original method.

### DIFF
--- a/config.go
+++ b/config.go
@@ -37,6 +37,8 @@ type Config struct {
 	HttpBasicAuthUser string
 	// http basic password
 	HttpBasicPassword string
+	// custom callback url
+	CallbackUrl string
 }
 
 // NewDefaultConfig create a default client config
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -48,5 +50,6 @@ func NewDefaultConfig() Config {
 		LogOutput:         ioutil.Discard,
 		HttpBasicAuthUser: "",
 		HttpBasicPassword: "",
-		RequestTimeout:    5}
+		RequestTimeout:    5,
+		CallbackUrl:       ""}
```
eicca: I believe that the nil value for string is `""` already, so it's not necessary to specify it.

config.go
```diff
@@ -37,6 +37,8 @@ type Config struct {
 	HttpBasicAuthUser string
 	// http basic password
 	HttpBasicPassword string
+	// custom callback url
+	CallbackUrl string
```
eicca: As far as I know in go convention it should be `CallbackURL`.

gambol99: yeah .. i think gofmt will complain as well ...

elliot:
Updated to meet code style

gambol99:
:+1: ... merging

PR #76, opened 2015-10-28
Fix examples code

x-cray:


examples/groups/main.go
```diff
@@ -106,7 +106,7 @@ func main() {
 	group := marathon.NewApplicationGroup(GROUP_NAME)
 	group.App(frontend).App(redis).App(mysql)
 
-	Assert(client.CreateGroup(group))
+	Assert(client.CreateGroup(group, true))
```
gambol99: https://github.com/gambol99/go-marathon/blob/master/group.go#L104

examples/applications/main.go
```diff
@@ -100,7 +100,7 @@ func main() {
 	glog.Infof("Successfully deleted the application")
 
 	glog.Infof("Starting the application again")
-	_, err = client.CreateApplication(application)
+	_, err = client.CreateApplication(application, true)
```
gambol99: https://github.com/gambol99/go-marathon/blob/master/application.go#L398

examples/groups/main.go
x-cray: Oh, sorry, it seems that I've been at wrong tree

PR #77, opened 2015-10-28
Implemented better error handling

Raffo:
I would love to discuss with you some improvement I've been doing to support a better error handling. I would love to discuss with you the following changes, which unfortunately are not shown here for GitHub limits. Please watch the changes on your local machine. 
The changes regard the following:

1) Introduced support for Godeps to enable repeatability of builds
2) Introduced a package with errors, similar to what is done in other golang projects (i.e. kubernetes)
3) Introduced different error handling types

The goal here is to provide to the code using this library the possibility to understand what is going on, reasoning on the type of error received. All the errors implement the "error" interface. 

I understand that this changes are quite big, but I need this to have a better error handling in my go code, with the previous implementation I could only print the string representation of the message which is not enough to implement any logic in code. 

I'm still working on this but I opened the PR for early discussion. 
PLEASE DO NOT MERGE BEFORE END OF DISCUSSION.


eicca:
>  Introduced support for Godeps to enable repeatability of builds

Can you please explain a bit more: why there's a need for "repeatability of builds" in the case of the library? Once an end user will install `go-marathon` with `go get` - the latest version of dependencies will be fetched anyway. And the downside is that `Godeps` folder will be fetched every time and won't be used. 
Please correct me if I'm wrong.

eicca:
>  Introduced a package with errors, similar to what is done in other golang projects (i.e. kubernetes)

I'm not sure if `go-marathon` could be comparable to the `kubernetes`. When `go-marathon` is a thin *library* wrapper around marathon HTTP interface, `kubernetes` is a sophisticated *application*. We can take another *library* example maintained by google https://github.com/grpc/grpc-go/blob/master/clientconn.go and they define errors in the similar way to `go-marathon`.

> The goal here is to provide to the code using this library the possibility to understand what is going on, reasoning on the type of error received.

Can you please explain this a bit more? I thought one receives `error` interface and then one can cast it. However in order to cast, one needs to know the type before.

> I understand that this changes are quite big, but I need this to have a better error handling in my go code, with the previous implementation I could only print the string representation of the message which is not enough to implement any logic in code.

```
type ErrorStatus struct {
	Message string //description of the error
	Status  string //one between "success" or "failure"
	Code    int    //http code, 0 if not set
}
```

I would assume that the status will be always `failure`, otherwise it doesn't make sense to return an `error`. And in my opinion one of the reasons of using a wrapper library around http api is to exactly abstract the details of the http codes and statuses.

Can you please provide an example in your code which needs the changes, so it will be easier to understand the use case and reason about it? Thanks :)

eicca:
I asuume that the diff for the error handling change could be seen here: https://github.com/Raffo/go-marathon/commit/b823f03bb48884daafe46a4555bf37b9fc38ed7f

Raffo:
@eicca First of all thank you for your comments. I started to do this on GH because I wanted early comments and it's important that you share what you think. I'll go point by point: 

>Can you please explain a bit more: why there's a need for "repeatability of builds" in the case of the >library? Once an end user will install go-marathon with go get - the latest version of dependencies will >be fetched anyway. And the downside is that Godeps folder will be fetched every time and won't be >used. 

That's exactly the point: with ```go get``` you get the latest version which means that if there's a change in a function definition everything will break. I saw this happen with other golang projects, that's why I think that Godeps could be used to fix that. It's probably not ideal, but as go-marathon has some external dependencies, it is one of the ways to go. 

>I'm not sure if go-marathon could be comparable to the kubernetes. When go-marathon is a thin >library wrapper around marathon HTTP interface, kubernetes is a sophisticated application. 

They're definitely not comparable, I was just mentioning a well written (according to me) golang project, I think it was clear I was not directly comparing a library with a huge project like kubernetes. 

>We can take another library example maintained by google https://github.com/grpc/grpc->go/blob/master/clientconn.go

Do you really think this is the way to go? I understand it is a library, but like this we are assuming that errors are basically only string descriptions of real errors that were instead coded with a lot more detail (i.e. HTTP error codes). If we only have a string we have no way to reason about the error in the code. Please have a look at how SyntaxError is defined in the json package (standard go, you can find an article here: https://blog.golang.org/error-handling-and-go ) 

> Can you please explain this a bit more? I thought one receives error interface and then one can cast > it. However in order to cast, one needs to know the type before.

That's again why I'm doing this open like that. I also don't think that's ideal, but this is what it is done in go standard libraries, referring to the same article linked before (https://blog.golang.org/error-handling-and-go): 

````
type SyntaxError struct {
    msg    string // description of error
    Offset int64  // error occurred after reading Offset bytes
}

func (e *SyntaxError) Error() string { return e.msg }
if err := dec.Decode(&val); err != nil {
    if serr, ok := err.(*json.SyntaxError); ok {
        line, col := findLine(f, serr.Offset)
        return fmt.Errorf("%s:%d:%d: %v", f.Name(), line, col, err)
    }
    return err
}
````

>I would assume that the status will be always failure, otherwise it doesn't make sense to return an >error. And in my opinion one of the reasons of using a wrapper library around http api is to exactly >abstract the details of the http codes and statuses.

Totally agree with both the points. I will definitely remove ```Status``` as it doesn't make any sense. 

Regarding ```Code``` I agree that can also be removed, but it wouldn't be bad to keep it to further inspect the error code returned for marathon as error codes for marathon are unfortunately not documented (at least not all of them) and must be looked inside the code. If we reach an agreement that we don't want this field because it's just useless I can remove it.  


>Can you please provide an example in your code which needs the changes, so it will be easier to >understand the use case and reason about it? Thanks :)

I can provide many of them if you need, but for the moment just think that I'm using go-marathon in a software that exposes itself a REST API and it's therefore important to know what exactly happened in the code itself, possibly without having to parse a string :-)

eicca:
Regarding the errors I checked several http wrapper libraries (I might be wrong, since it was just a quick look):
- kubernetes library just returns errors https://github.com/kubernetes/kubernetes/blob/master/pkg/client/unversioned/request.go#L689 from the request or fmt.Errorf.
- consul api client returns result, meta information and plain errors. However meta information doesn't include errors info https://github.com/hashicorp/consul/tree/master/api
- github api wrapper https://github.com/google/go-github/blob/master/github/github.go#L300 returns result, response and error. Here the codes and information could be taken from the response. I'm not sure how often it's used.
- I found the similar thing in https://github.com/GoogleCloudPlatform/gcloud-golang/blob/master/bigquery/error.go (note: from the `gcloud-golang` looks like only bigquery has it). The fields look quite specific to the bigquery though.

I'm not sure what would be the best practice here. I might assume that the REST wrapper around another REST wrapper (go-marathon) which needs the specifics of response might be too specific case for it. 

>  If we only have a string we have no way to reason about the error in the code. 
We don't have only string, the errors are defined as variables, so you can compare them with what you get from a method call.

eicca:
Regarding the godeps:
I checked the commited dependencies and `go-marathon` depends only on `glog` which is highly unlikely to change and `testify` for tests. In my opinion `testify` is not worth it and I would even think of getting rid of this library, since the common practice (as far as I know) is to write table tests or write assertions yourself.

Raffo:
>I'm not sure what would be the best practice here. I might assume that the REST wrapper around >another REST wrapper (go-marathon) which needs the specifics of response might be too specific case >for it.

Again, we don't really need error codes, I agree. 

>We don't have only string, the errors are defined as variables, so you can compare them with what you >get from a method call.

I see what you mean here, I didn't think about comparing directly the variables and I don't find it particularly elegant, but it might work. 

>Regarding the godeps:
>I checked the commited dependencies and go-marathon depends only on glog which is highly unlikely >to change and testify for tests. In my opinion testify is not worth it and I would even think of getting rid of >this library, since the common practice (as far as I know) is to write table tests or write assertions >yourself.

I honestly don't understand why you are so against godeps "It's highly unlikely" is not enough to me, and Godeps makes just the repo bigger which I don't see it as a problem in the year 2015. And once it's there you can still build using your plain go dependencies. I find it needed and definitely not a problem if it's there. I could think about removing all the libraries committed, there should be an option for that, but I don't find this so important.

Raffo:
I think that comparing the variables could be enough and it could be used to avoid changing the source code as I had to do for this PR. I would be in favor of this approach, if you agree. I would love to do the following:

1) Improving the checked error codes to reflect the real ones in the API. 
2) Keeping the Godeps, possibly without committing the dependencies if we agree we don't want them. 

This means closing this PR and create another one with the changes.

Raffo:
Having a deeper look at the ```apiCall``` function:

````
func (r *marathonClient) apiCall(method, uri, body string, result interface{}) (int, string, error) {
	glog.V(DEBUG_LEVEL).Infof("[api]: method: %s, uri: %s, body: %s", method, uri, body)

	status, content, _, err := r.httpRequest(method, uri, body)
	if err != nil {
		return 0, "", err
	}

	glog.V(DEBUG_LEVEL).Infof("[api] result: status: %d, content: %s\n", status, content)
	if status >= 200 && status <= 299 {
		if result != nil {
			if err := r.decodeRequest(strings.NewReader(content), result); err != nil {
				glog.V(DEBUG_LEVEL).Infof("failed to unmarshall the response from marathon, error: %s", err)
				return status, content, ErrInvalidResponse
			}
		}
		return status, content, nil
	}

	switch status {
	case 500:
		return status, "", ErrInvalidResponse
	case 404:
		return status, "", ErrDoesNotExist
	}

	// step: lets decode into a error message
	var message struct {
		Message string `json:"message"`
	}

	if err := r.decodeRequest(strings.NewReader(content), &message); err != nil {
		return status, content, ErrInvalidResponse
	}

	errorMessage := "unknown error"
	if message.Message != "" {
		errorMessage = message.Message
	}

	return status, "", fmt.Errorf("%s", errorMessage)
}
````

Here the problems I see:
1. the switch on status is done before reading the message (maybe no message is returned), but only two error codes are tested among the ones I posted above
2. when the message is decoded, then a simple message is returned without taking care of the tipe of error, not using any of the preallocated variables and so giving no way to a user of the code of checking the type (or the variable).
3. The only function using ```apiCall``` are in client.go and group.go. In the first, the first two returned values are ignored (using underscores), in the second only the status is used and this is what is done:
````
func (r *marathonClient) HasGroup(name string) (bool, error) {
	uri := fmt.Sprintf("%s/%s", MARATHON_API_GROUPS, trimRootPath(name))
	status, _, err := r.apiCall("GET", uri, "", nil)
	if err == nil {
		return true, nil
	}
	if status == 404 {
		return false, nil
	}

	return false, err
}
````
In this case using a struct as I did would be much more useful. 

What I see is that variables are clearly very good when we're returning an error that is generated from ```go-marathon``` itself, as we are the one creating the check. When interacting with ```marathon``` REST API the thing need to be changed if we want to have a better error handling. 

@eicca I'm waiting for your comments :-)

eicca:
@Raffo I think it would make sense to create small pull requests with the changes you want to make, so it would be a bit easier to discuss and to approach. I totally agree that we can improve the error handling here, since we're missing a lot of the errors. Additionally we can refactor the api call part.

Raffo:
@eicca I agree, I'll try to do it. My previous point was only that we should probably keep structures instead of variables to encapsulate the messages coming from the backend without relying on a simple generic error. If so, the PR could be mostly similar to what I did here... I can drop godeps and try to redo this in a better way, if you agree... but I'll have to keep the structs.

eicca:
Sorry for the late response. I'm sorry, I couldn't understand well the proposed changes after our discussion since we're talking about few things at the same time :) that's why I thought that it would be easier to understand after dividing them.

Maybe the structs will be better and will simplify the code. Maybe it's better to return the whole response from the `apiCall` and then take body and status out if it if needed and return the errors as variables.

I'm a bit puzzled about the intention, that's why maybe I'm missing the point (I'm really sorry for this). Is the intention:
a) to return status code of marathon api to end user
b) to refactor the internals by changing the inner logic of the status handling
c) to return more different errors then currently we do

@gambol99 maybe you can see this clearer and can tell your opinion?

Thanks :)

Raffo:
@eicca No problem for the late reply, I was very busy too and I didn't have time to do this anyway :-) 

I would say b and c: I want to refactor the internals as error handling is not consistent for check done by go-marathon (i.e. parse error) and error returned by the marathon api. From an end user perspective there's no reason that this should be like that. 
I will create a better pull request and close this one, to simplify the whole thing.

eicca:
awesome :+1:

PR #79, opened 2015-10-28
Implement SSE events transport. Add new event types. Update README

### LINKED ISSUE #19: Unreliable IP retrieval for Event Subscription
client.RegisterSubscription in subscription.go assumes that the interface passed in from the config is unique. On my (and I'm sure most) machines, there is are  IPv6 and IPv4  interfaces with the same name. The matching in the getInterfaceAddress in utils.go will most likely get an unusable address. I'm not sure what the best way to handle this is though :(. Perhaps optionally allow an IP to be explicitly set in the config which puts the onus on the developer to retrieve and use the correct IP?

x-cray:
Kind of fixes #19
Please review the PR, I'm a bit new to Go


eicca:
As current code doesn't fully pass `golint`, would it make sense to write the new code in a proper way, or in a consistent way? Writing the golintable code would make it easier to refactor the old code in the future, however it will look inconsistent. What do you think, what's better?

README.md
```diff
@@ -105,52 +104,73 @@ if err := client.ScaleApplicationInstances(application.ID, 10); err != nil {
...
-	    glog.Infof("EVENT: %s", event )
+	// lets register for events
+	update := make(marathon.EventsChannel, 5)
+	if err := client.AddEventsListener(update, marathon.EVENTS_APPLICATIONS); err != nil {
+		glog.Fatalf("Failed to register for subscriptions, %s", err)
+	} else {
+		for {
+			event := <-update
+			glog.Infof("EVENT: %s", event)
+		}
```
eicca: Since you're touching the examples anyway, what do you think about removing `else` statements from them?

subscription.go
```diff
@@ -140,6 +152,35 @@ func (r *marathonClient) RegisterSubscription() error {
...
+	if err != nil {
+		return err
+	}
+	url := fmt.Sprintf("%s/%s", marathon, MARATHON_API_EVENT_STREAM)
+	stream, err := eventsource.Subscribe(url, "")
+	if err != nil {
+		glog.V(DEBUG_LEVEL).Infof("failed to connect to Marathon event stream, error: %s", err)
+		r.cluster.MarkDown()
+		// step: retry the request with another endpoint
+		return r.registerSSESubscription()
```
eicca: What do you think about converting recursion to the loop? It might be a bit easier to read and to understand.

eicca:
I wonder if it's possible to have some tests for this addition. Do you have any thoughts about this?

x-cray:
@eicca As I mentioned earlier, I'm new to Go =) I've checked golint output and it seems that there are many places it complains about. I'll fix points which are related to this PR, but IMO others should be fixed is separate PRs to maintain consistent intent of this specific PR.

subscription.go
x-cray: Yes, I thought about that. This piece of code was mainly copied from https://github.com/gambol99/go-marathon/blob/master/client.go#L334 which maybe is also a subject for refactoring.

README.md
x-cray: Ok, I'll also modify actual example code file.

x-cray:
@eicca In terms of tests I think that might mock SSE server like in this example https://github.com/donovanhide/eventsource/blob/master/example_event_test.go and sent predefined set of events expecting to receive them from this channel https://github.com/gambol99/go-marathon/blob/master/subscription.go#L47

client.go
```diff
@@ -304,34 +306,39 @@ func (r *marathonClient) apiCall(method, uri, body string, result interface{}) (
...
+	var response *http.Response
 
-	// step: get a member from the cluster
-	marathon, err := r.cluster.GetMember()
-	if err != nil {
-		return 0, "", nil, err
-	}
+	// Try to connect to Marathon until succeed or
+	// the whole custer is down
+	for {
```
gambol99: :+1: definitely easier to read when using the for loop ...

gambol99:
:+1: LGTM

x-cray:
@gambol99 I'd like to add some tests to it.

subscription.go
```diff
@@ -140,6 +154,49 @@ func (r *marathonClient) RegisterSubscription() error {
 	return nil
 }
 
+func (r *marathonClient) registerSSESubscription() error {
+	// Prevent multiple SSE subscriptions
+	if r.subscribedToSSE {
+		return nil
+	}
```
eicca: It's still could be affected by a race condition. Would it be better to try to use https://golang.org/pkg/sync/#Once? Or that would be totally users fault to achieve the race condition here and the boolean field is enough? :)

x-cray: @eicca Good point. But I see another issue here: `registerSSESubscription()` is only called from within `RegisterSubscription()` which in turn is *only* called from `AddEventsListener()` which is already synchronised. The issue is that `RegisterSubscription` itself is being exported while it is not defined in `Marathon` interface. I think that we could just rename `RegisterSubscription` to `registerSubscription` and that's it.

x-cray:
@gambol99 @eicca Question: should `Marathon.UnSubscribe()` be exported to interface? It is being called automatically upon removing event listener in `Marathon.RemoveEventsListener()`. Would it be better is we don't export it? What's the purpose of this method if we call it from outside while having some event subscriptions active?

eicca:
@x-cray I'm sorry for the lack of activity in this PR, unfortunately I don't have enough knowledge of the marathon events. I can try to dig into it a bit later once I have more time.

gambol99:
@x-cray good spot! and no under this context it shouldn't really be exported, neither should the RegisterSubscription() from the looks for it, as the {Add,Remove}EventsListener() methods can handled it all. Off the top of my head though, it might be useful to keep the method but change its function slightly

```go
from
func (r *marathonClient) UnSubscribe() error 
to 
func (r *marathonClient) UnSubscribe(url string) error
```
And then in https://github.com/gambol99/go-marathon/blob/master/subscription.go#L66 where it checks for any more listeners we could change to 
```go
if len(r.listeners) <= 0 {
	r.UnSubscriber(r.SubscriptionURL())
}
```  
Internally the consumer should use the Add/RemoveEventslisten(channel) to handle events, but they can also have a generic method to delete subscriptions(url) if they like, i.e. list all and clean up subscriptions unrelated to the Add/Register methods etc ... what you reckon? ..

x-cray:
@gambol99 In my opinion your abstraction over Marathon events API with `{Add/Remove}EventsListener()` is really great and it makes `UnSubscribe()` unnecessary. However, your suggestion to add `url` param adds some sense to it, but only in situation when we need to unsubscribe some third-party subscriber. Could be. Will you make this change in upstream?

gambol99:
@x-cray 
> However, your suggestion to add url param adds some sense to it, but only in situation when we
> need to unsubscribe some third-party subscriber

agreed .. 

> Will you make this change in upstream?

If this PR doesn't need to change it i'm happy to raise the change and PR myself? ...

x-cray:
> If this PR doesn't need to change it i'm happy to raise the change and PR myself? ...
I think that it's not related to PR. I've only changed [this line](https://github.com/x-cray/go-marathon/blob/d4fc9f243710e423e548176fb773621b6697995c/subscription.go#L76) by adding check for correct `EventsTransport` to unsubscribe only when using callback event transport.

For the rest, I'm happy with all my changes now =) I've added functional tests for event stream. The most notable change in unit testing runtime infrastructure is the way how test HTTP server is started now — for every test case a new instance is being started and stopped afterwards. That was required for event stream testing and in general it's always good practice to prepare test environment from scratch for every distinct test and then dispose it.

testing_utils.go
```diff
@@ -0,0 +1,169 @@
...
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/donovanhide/eventsource"
+	"github.com/golang/glog"
+	yaml "gopkg.in/yaml.v2"
+	"sync"
```
gambol99: 'sync' should probably be placed in the above section

gambol99:
Aside from the one tiny comment ... LGTM ... awesome PR and cheers for fixing up alot of the linting issues

testing_utils.go
x-cray: Sure, missed that one

Issue #81, opened 2015-10-30
Filter getTasks by status


emilevauge:
Hi!
Marathon's API allows to filter by status when retrieving tasks:
https://mesosphere.github.io/marathon/docs/rest-api.html#get-v2-tasks
It would be great to add that possibility here https://github.com/gambol99/go-marathon/blob/master/task.go#L59 by adding a parameter `status string` to the method `func (r *marathonClient) AllTasks() (*Tasks, error)`.
For now, it's impossible to get tasks' status with `gambol99/go-marathon`.

eicca:
Makes sense!
It should also affect `ListTasks` https://github.com/gambol99/go-marathon/blob/master/task.go#L80 I think. To be honest I would prefer to get rid of the `ListTasks` as it's just a simple data manipulation.

eicca:
@emilevauge I just pushed the changes to master. Can you please check if it solves the issue now?

emilevauge:
@eicca, great! I Will test that on [traefik](http://github.com/emilevauge/traefik) ASAP :)

gambol99:
closing the issue as solved?

emilevauge:
Yep :) Solved. Thanks!

PR #82, opened 2015-10-31
AllTasks handles status parameter

eicca:
Resolves https://github.com/gambol99/go-marathon/issues/81

As it was mentioned in https://github.com/gambol99/go-marathon/issues/81 the `ListTasks()` is gone. What do you think about it?

### DIFF
--- a/client.go
+++ b/client.go
@@ -72,9 +72,7 @@ type Marathon interface {
 	// get a list of tasks for a specific application
 	Tasks(application string) (*Tasks, error)
 	// get a list of all tasks
-	AllTasks() (*Tasks, error)
-	// get a listing of the task ids
-	ListTasks() ([]string, error)
+	AllTasks(status string) (*Tasks, error)
 	// get the endpoints for a service on a application
 	TaskEndpoints(name string, port int, healthCheck bool) ([]string, error)
 	// kill all the tasks for any application
--- a/client.go
+++ b/client.go
@@ -72,7 +72,7 @@ type Marathon interface {
 	// get a list of tasks for a specific application
 	Tasks(application string) (*Tasks, error)
 	// get a list of all tasks
-	AllTasks(status string) (*Tasks, error)
+	AllTasks(v url.Values) (*Tasks, error)
 	// get the endpoints for a service on a application
 	TaskEndpoints(name string, port int, healthCheck bool) ([]string, error)
 	// kill all the tasks for any application
--- a/task_test.go
+++ b/task_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package marathon
 
 import (
+	"net/url"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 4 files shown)


client.go
```diff
@@ -72,9 +72,7 @@ type Marathon interface {
 	// get a list of tasks for a specific application
 	Tasks(application string) (*Tasks, error)
 	// get a list of all tasks
-	AllTasks() (*Tasks, error)
-	// get a listing of the task ids
-	ListTasks() ([]string, error)
+	AllTasks(status string) (*Tasks, error)
```
gambol99: Perhaps we could use url.Values (https://github.com/gambol99/go-marathon/blob/master/application.go#L265), the benefit being we don't have to change anything if or when the API adds another option ... what you reckon?

Raffo: I agree with url.Values. The PR seems good for the rest and it's a very useful feature.

eicca: I think I would rather help a user by explicitly providing possible parameters. In this case you don't need to look to the api and you have compile time check that your params are right.
There are two downsides with this approach:
1) as it was mentioned: api might change
2) providing nil values will be ugly

We can solve `2)` by wrapping request params into structs, and I think it might look actually nice.

I think at this point I will do `url.Values` and then we can decide if we want to refactor the way parameters are passed.

task_test.go
```diff
@@ -24,12 +25,20 @@ import (
...
-	tasks, err := client.AllTasks()
+	tasks, err := client.AllTasks(nil)
 	assert.Nil(t, err)
 	assert.NotNil(t, tasks)
 	assert.Equal(t, len(tasks.Tasks), 2)
 }
 
+func TestAllStagingTasks(t *testing.T) {
+	client := NewFakeMarathonEndpoint(t)
+	tasks, err := client.AllTasks(url.Values{"status": []string{"staging"}})
```
eicca: > I think I would rather help a user by explicitly providing possible parameters. In this case you don't need to look to the api and you have compile time check that your params are right.
There are two downsides with this approach:
1) as it was mentioned: api might change
2) providing nil values will be ugly
We can solve 2) by wrapping request params into structs, and I think it might look actually nice.
I think at this point I will do url.Values and then we can decide if we want to refactor the way parameters are passed.

So my suggestion would be something like:
```
client.AllTasks(AllTasksRequest{status: marathon.TaskStaging})
```

Or since marathon doesn't have more then 2 parameters the first version even could be ok:
```
client.AllTasks(marathon.TaskAny)
// or 
client.AllTasks(maraton.TaskStaging)
```
Again, it's just my preference :)

gambol99: @eicca sorry for the delay ... I guess my reasoning behind I didn't fancy chasing the API. Side stepping just AllTasks() for moment has one; starts off with one variable, then another comes along, so you add it and break it all other builds (yes, we have godeps for that, but), then perhaps the second option is supported in one version but not the early release's etc etc .. and then times by x methods, i've a feeling it could get messy. Wrapping them in structs does mitigate most of that, it's simply a question if you want those pretty (e.g marathon.TaskStaging) or opened up by using a url.Values and letting the user decide ...

eicca: I agree. As it was said let's stick with `url.Values`. We can change it in the future if we realize that we need another approach.

gambol99:
@eicca LGTM

PR #86, opened 2015-11-05
unsubscribe url parameter

gambol99:
summary: the current method doesn't really make sense, adding a param allows us to control the callback param as mentioned #79 

- changing the unsubscribe method to accept a url parameter
- adding a golint to the makefile

### DIFF
--- a/client.go
+++ b/client.go
@@ -119,7 +119,7 @@ type Marathon interface {
 	// remove a events listener
 	RemoveEventsListener(channel EventsChannel)
 	// remove our self from subscriptions
-	UnSubscribe() error
+	Unsubscribe(string) error
 
 	// --- MISC ---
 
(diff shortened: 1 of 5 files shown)


client.go
```diff
@@ -119,7 +119,7 @@ type Marathon interface {
 	// remove a events listener
 	RemoveEventsListener(channel EventsChannel)
 	// remove our self from subscriptions
-	UnSubscribe() error
+	UnSubscribe(string) error
```
eicca: Since we are changing the function anyway: would you like to rename it to `Unsubscribe`? :) http://www.thefreedictionary.com/unsubscribe
(sorry for this type of comment) :)

gambol99:
@eicca the comment is welcome :-) ... and yep, it makes sense, I'll change the method name

gambol99:
renamed the method and rebased

eicca:
:+1:

PR #87, opened 2015-11-06
Add Application.AddLabel method.

timoreimann:
Similar to AddEnv, the method makes sure that the Labels map is initialized before adding an item.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,8 @@ application.CPU(0.1).Memory(64).Storage(0.0).Count(2)
 application.Arg("/usr/sbin/apache2ctl").Arg("-D").Arg("FOREGROUND")
 application.AddEnv("NAME", "frontend_http")
 application.AddEnv("SERVICE_80_NAME", "test_http")
+application.AddLabel("environment", "staging")
+application.AddLabel("security", "none")
 // add the docker container
 application.Container.Docker.Container("quay.io/gambol99/apache-php:latest").Expose(80).Expose(443)
 application.CheckHTTP("/health", 10, 5)
--- a/application.go
+++ b/application.go
@@ -182,6 +182,18 @@ func (r *Application) AddEnv(name, value string) *Application {
 	return r
 }
 
+// AddLabel adds a label to the application
+//		name:	the name of the label
+//		value:	go figure, the value associated to the above
+func (r *Application) AddLabel(name, value string) *Application {
+	if r.Labels == nil {
+		r.Labels = make(map[string]string, 0)
+	}
+	r.Labels[name] = value
+
+	return r
+}
+
 // HasHealthChecks is more of a helper method, used to check if an application has healtchecks
 func (r *Application) HasHealthChecks() bool {
 	if r.HealthChecks != nil && len(r.HealthChecks) > 0 {
--- a/application_test.go
+++ b/application_test.go
@@ -80,6 +80,13 @@ func TestApplicationEnvs(t *testing.T) {
 	assert.Equal(t, 1, len(app.Env))
 }
 
+func TestApplicationLabels(t *testing.T) {
+	app := NewDockerApplication()
+	assert.Equal(t, 0, len(app.Labels))
+	app.AddLabel("hello", "world")
+	assert.Equal(t, 1, len(app.Labels))
+}
+
 func TestHasHealthChecks(t *testing.T) {
 	app := NewDockerApplication()
 	assert.False(t, app.HasHealthChecks())


eicca:
As the change looks consistent to what the `AddEnv` does, in my opinion it would be better to remove the *smart* setters and initialize the structs inside of the constructor. Examples:
- http://stackoverflow.com/questions/4498998/how-to-initialize-members-in-go-struct
- http://stackoverflow.com/questions/27553399/golang-how-to-initialize-a-map-field-within-a-struct

In this case I would remove the similar setter and modify `NewDockerApplication` constructor.

Also it could make sense to remove getters as well since the variables are exported and there's almost no transformation logic (in another PR).

@gambol99 what do you think about it?

timoreimann:
To me, it boils down to the question of how DSL-ish go-marathon wants to be: Most application fields come with setters to allow for a fluent usage; however, a few are missed out (like `Labels` and `User`). Was this intentional, and if so, what's the criteria under which a setter is added or not? On the contrary, if they were really just overlooked, I'd opt to have the `AddLabel` convenience setter for consisteny and usability reasons.

Removing existing methods would imply a breaking API change. Does the project follow a particular policy regarding API stability?

@eicca, you mention removing getters too. Which ones do you mean exactly? At least for `Env` and `Labels`,  there doesn't seem to be any getters.

Happy to implement whatever approach we decide to take on!

eicca:
> Removing existing methods would imply a breaking API change. Does the project follow a particular policy regarding API stability?

It's a good point. I assume that the project is in early state since we're breaking API almost in every second PR. Maybe we can mention this in the README. I think it will stabilize soon together with Marathon.

> @eicca, you mention removing getters too. Which ones do you mean exactly? At least for Env and Labels, there doesn't seem to be any getters.

Oh, I'm sorry, I was wrong, there are only setters of course. I was confused because of the convention https://golang.org/doc/effective_go.html#Getters

> To me, it boils down to the question of how DSL-ish go-marathon wants to be: Most application fields come with setters to allow for a fluent usage; however, a few are missed out (like Labels and User). Was this intentional, and if so, what's the criteria under which a setter is added or not? On the contrary, if they were really just overlooked/not implemented yet, I'd opt to have the AddLabel setter for consistency and usability reasons (and support fully DSLifyng the API).

I would ask what @gambol99 thinks about it. In my opinion there's no need for getters or setters if we operate just on golang data structures. As I see it the library has the knowledge of how to communicate with marathon and represent data in the golang + additional helpers related to the marathon knowledge. Also having one way to do the thing might confuse the end user less.

gambol99:
I'm probably somewhat biased on this; as I spent a large amount of time writing ruby :-) ... Admittedly, removing those setters would a pretty big break and given all the fields are exported, as is, the decision is in the hands of the developer as to which they prefer.

```go
   marathon.Application{
		ID: "myapplication",
		CPUs: 0.1,
		Mem: 64,
		Instances: 2,
		Args: []string{"usr/sbin/apache2ctl","-D","FOREGROUND"},
		Env: map[string]string{
			"NAME" : "frontend_http",
			"SERVICE_80_NAME" : "test_http",
		},
		RequirePorts: true,
		Type: "DOCKER",
		Container: &marathon.Docker{
			Image:  "apache",
			Network:      "BRIDGE",
			PortMappings: []marathon.PortMapping{
				{
					Protocol: "TCP",
					ContainerPort: 80,
				},
				{
					Protocol: "TCP",
					ContainerPort: 443,
				},
			},
		},
		HealthChecks: []marathon.HealthCheck{
			{
				Protocol:               "HTTP",
				Path:                   "/health",
				GracePeriodSeconds:     30,
				IntervalSeconds:        10,
				MaxConsecutiveFailures: 3,
				TimeoutSeconds:         5,
			},
		},
	}
```

```go 
        app := marathon.NewDockerApplication()
	app.Name(applicationName).CPU(0.1).Memory(64).Count(2)
	app.Arg("/usr/sbin/apache2ctl").Arg("-D").Arg("FOREGROUND")
	app.AddEnv("NAME", "frontend_http").AddEnv("SERVICE_80_NAME", "test_http")
	app.RequirePorts = true
	app.Container.Docker.Container("apache").Expose(80).Expose(443)
	app.CheckHTTP("/heath", 80, 5)
```

eicca:
@gambol99 I see the point now. In this example the DSL way looks much better.
Sorry for bringing confusion :)

eicca:
LGTM

timoreimann:
I am with @gambol99 regarding the value of the DSL allowing for concise definitions of Marathon applications; his example makes this pretty clear in my opinion. (Although it's somewhat biased towards the case where the DSL shines, which is creating the fairly complex Docker application and using mostly default parameters.)

On the other hand, I agree with @eicca that having two ways to define applications is a bit confusing on first sight. I personally had to do some code-digging initially to understand that the DSL is "just" for convenience reasons. My plan is to follow up with a PR against the README file to add some clarification in this regards.

Overall, I'm in favor of the DSL approach though due to the level of verbosity required for explicit struct initialization and the cases where the DSL enables multiple initializations under the hood (as in a Docker application).

I forgot to extend the README file by an example of the new `AddLabel` method, so I just re-pushed onto by branch. Is there further need for (design) discussions, or do we think the PR is now good to merge?

gambol99:
:+1:

PR #88, opened 2015-11-07
Add README note re: API stability.

timoreimann:
This is a follow-up of a discussion that took place along PR #87.


gambol99:
:+1:

PR #89, opened 2015-11-07
Add note to README re: go-marathon's DSL purpose.

timoreimann:
This is a follow-up of a discussion that took place along PR #87.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,8 @@ if _, err := client.CreateApplication(application); err != nil {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


eicca:
What do you think about providing a code example as well? Some people (like me) skip explanations and look only at code examples :)

timoreimann:
To me, the note is mostly of educational nature. Thus, I'm a bit worried that embedding an entire comparative example could add some extra noise to the README file we might want to avoid.

How about just linking to @gambol99's excellent example in [his PR comment](https://github.com/gambol99/go-marathon/pull/87#issuecomment-154740328) instead? WDYT?

eicca:
I like the idea, however the linked example might become outdated, since it's not the part of the project. Let's keep it without changes then.
LGTM

PR #90, opened 2015-11-07
Make Application.{Arg,DependsOn}, Docker.Expose* variadic.

timoreimann:
The following unary functions can be invoked multiple times to
effectively set a collection of values:

- `Application.Arg()`
- `Application.DependsOn()`
- `Docker.Expose()`
- `Docker.ExposeUDP()`

This change makes the before-mentioned functions variadic so that
multiple values can be passed more conveniently.

Note that `Arg()` should actually be called `Args()` now. This requires a
bit more refactoring, however, as the struct receiver the method is
defined on already holds a field by the same name. In order to keep the
change set manageable, we'll leave the task of renaming the method to a
separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,11 +79,11 @@ glog.Infof("Deploying a new application")
 application := marathon.NewDockerApplication()
 application.Name("/product/name/frontend")
 application.CPU(0.1).Memory(64).Storage(0.0).Count(2)
-application.Arg("/usr/sbin/apache2ctl").Arg("-D").Arg("FOREGROUND")
+application.Arg("/usr/sbin/apache2ctl", "-D", "FOREGROUND")
 application.AddEnv("NAME", "frontend_http")
 application.AddEnv("SERVICE_80_NAME", "test_http")
 // add the docker container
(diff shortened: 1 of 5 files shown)


eicca:
LGTM

>Note that Arg() should actually be called Args() now. This requires a
bit more refactoring, however, as the struct receiver the method is
defined on already holds a field by the same name. In order to keep the
change set manageable, we'll leave the task of renaming the method to a
separate PR.

Maybe we can rename setters according to the convention (I already mentioned it in the previous PR):
https://golang.org/doc/effective_go.html#Getters

timoreimann:
Using the official naming convention sounds like a good idea to me. Especially since it will allow us to use names that match Marathon's domain language (e.g., `Name()` can be `SetAppID()` then).

timoreimann:
To avoid any misunderstandings: IMHO, the renaming should happen in a separate PR. Thus, I consider this PR to be ready for merging.

Please LMK if you think something is still missing. Thanks.

gambol99:
LGTM

PR #91, opened 2015-11-08
Make TestUtilsGetInterfaceAddress platform-independent.

timoreimann:
`TestUtilsGetInterfaceAddress()` in `utils_test.go` relies on the availability of a `lo` interface containing `127.0.0.1` as its first address member. At least on Mac OS X, this does not work because the
localhost device is named `lo0` and first address is the IPv6 one (`::1`).

The existing test is modified to first determine a valid address and subsequently runs the test with the expectation set against the found address. Along the way, we break out and test a helper function needed by both the implementation and the updated `TestUtilsGetInterfaceAddress()` test.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -87,7 +87,7 @@ func getInterfaceAddress(name string) (string, error) {
 			}
 			// step: return the first address
 			if len(addrs) > 0 {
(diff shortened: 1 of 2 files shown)


gambol99:
LGMT

PR #92, opened 2015-11-10
Remove Task.String()

mattes:
Either String() includes all Task fields or we just remove it ;-) I think the latter option just makes it easier to maintain in general.


eicca:
:+1:

gambol99:
LGTM

Issue #80, opened 2015-10-29
Golint conformance


x-cray:
I'm opening this issue here for discussion on the subject. @eicca suggested (https://github.com/gambol99/go-marathon/pull/79#issuecomment-152015182) that we might consider making project code meet golint rules. However, there are many places which are not just simply missing doc comments but require renaming of exported names. This will definitely be a breaking change. So, two questions here:
* Do we introduce a versioning strategy? (Also to track Marathon API changes)
* Do we try to make code meet golint rules?

eicca:
> Do we introduce a versioning strategy? (Also to track Marathon API changes)
I think at this stage it would be nice to have some versioning and to have a simple changelog at least.

> Do we try to make code meet golint rules?
In my opinion it definitely makes sense. I started making changes some time ago here: https://github.com/eicca/go-marathon/commit/cc88c7f66dc2e2a167aae98183a22af2894187df but then I realized that there's much more to change, so I got lazy a bit :/

eicca:
There are two things left:
- events.go
- eventsource `Event.Id()` method

PR #93, opened 2015-11-10
Apply style changes

eicca:
Part of #80.
I didn't lint events, since linted constant name would be the same as struct name. Maybe we need to refactor the events part. Unfortunately I didn't use events, so I lack knowledge about them.

Also `Event.Id()` couldn't be linted, since the eventsource library also is not linted.


config.go
```diff
@@ -54,8 +54,8 @@ func NewDefaultConfig() Config {
 		EventsPort:        10001,
 		EventsInterface:   "eth0",
 		LogOutput:         ioutil.Discard,
-		HttpBasicAuthUser: "",
-		HttpBasicPassword: "",
+		HTTPBasicAuthUser: "",
```
gambol99: Since were touching it ... should we remove the defaulted fields? ...  i.e. HTTPBasicAuthUser, HTTPBasicPassword and CallbackURL? ... all of which would be empty anyhow ...

gambol99:
other than the one comment .. LGTM ... squash and merge :-)

config.go
eicca: Yes, makes sense.

Issue #66, opened 2015-10-12
generic logger 


gambol99:
Implement a generic logger interface, defaulting to glog but allowing the consumer to override and run their own ..

x-cray:
BTW, how `config.LogOutput` is actually used?

eicca:
I assume it's not used any more. All logging is done with `glog` currently which is a bit inflexible.

x-cray:
Ok, I removed its assignment from examples code in #79

gambol99:
@x-cray, it was left over from the previous logging code ... although this issue is supposed act as a reminder to replace that as well :-) ..

mattes:
We are using a customized glog version. As soon as we include another package that imports the "original" glog version, we get a panic: 

```
panic:
xxx.test flag redefined: log_dir
```

... which makes sense. Not sure how to solve this, yet.

mattes:
Maybe this helps: https://github.com/golang/glog/blob/master/glog.go#L914

eicca:
@mattes I think the best solution would be to not to use `glog` directly, rather injecting logger to the library through config.
I know that it was done in `grpc-go` for example: https://github.com/grpc/grpc-go/tree/master/grpclog
We can do the same, though maybe there are some better practices?

mattes:
ref: #103

PR #94, opened 2015-11-13
remove LogOutput as it's not being used anywhere

mattes:
maybe add a note to use -v=10 with glog to get logging


timoreimann:
Also mentioned in #66

eicca:
LGTM

gambol99:
LGTM :+1:

PR #96, opened 2015-11-13
Add ErrConflict

mattes:
This is needed for `func (r *marathonClient) CreateApplication(application *Application) (*Application, error)`

https://github.com/gambol99/go-marathon/blob/master/application.go#L410

### DIFF
--- a/client.go
+++ b/client.go
@@ -148,6 +148,8 @@ var (
 	ErrInvalidArgument = errors.New("the argument passed is invalid")
 	// ErrTimeoutError is thrown when the operation has timed out
 	ErrTimeoutError = errors.New("the operation has timed out")
+	// ErrConflict is thrown when the resource is conflicting (ie. app already exists)
+	ErrConflict = errors.New("conflicting resource")
 )
 
 type marathonClient struct {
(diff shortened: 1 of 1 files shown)


timoreimann:
When Marathon returns 409, it adds further error details to the response message. (See the 409 error response example for the [PUT](https://mesosphere.github.io/marathon/docs/rest-api.html#put-v2-apps-appid) and the [POST request](https://mesosphere.github.io/marathon/docs/rest-api.html#post-v2-apps), respectively.) It would be quite helpful to expose this information to the user.

We could amend the PR, or create another issue to track future effort in this direction.

eicca:
Related to https://github.com/gambol99/go-marathon/pull/77

gambol99:
sorry for the delay ... This would be that the error message decoded at https://github.com/mattes/go-marathon/blob/patch-2/client.go#L297 would no longer be retrieved ... Admittedly though since its passes nothing but a generic error and string its unlucky people are using it effectively. Would be nice to have a Error per response type? ... Either, i'll chunk that into a issue for later and merge ... anyone else see a problem with this?

timoreimann:
I just created and self-commented #101 where I am proposing a slightly different general approach to how and what kinds of errors we should expose. Happy to hear your thoughts.

PR #103, opened 2015-11-21
Replaced glog with pkg/log and some housekeeping

mattes:
* Replace glog with pkg/log
* Added an glog example app
* No multiline logging anymore (which makes it easier for anything that tries to parse log files)
* Removed random and inconsistent logging. Don't log in functions that return an error anyway for example. Let the user decide what to do with the error. 
* Cleaned up apiOperation/ apiCall /httpRequest
* Updated `func (r *marathonCluster) GetMember()` to call the `MarkDown` func. This separates concerns, and we don't have to worry about it in `httpRequest()` or in `registerSSESubscription()`

Can someone give me some feedback? :-)

### DIFF
--- a/application.go
+++ b/application.go
@@ -19,7 +19,6 @@ package marathon
 import (
 	"errors"
 	"fmt"
-	"log"
 	"net/url"
 	"time"
 )
@@ -479,7 +478,6 @@ func (r *marathonClient) DeleteApplication(name string) (*DeploymentID, error) {
 // RestartApplication performs a rolling restart of marathon application
 // 		name: 		the id used to identify the application
 func (r *marathonClient) RestartApplication(name string, force bool) (*DeploymentID, error) {
-	log.Printf("restarting the application: %s, force: %v", name, force)
 	deployment := new(DeploymentID)
 	var options struct {
 		Force bool `json:"force"`
@@ -497,7 +495,6 @@ func (r *marathonClient) RestartApplication(name string, force bool) (*Deploymen
 // 		instances:	the number of instances you wish to change to
 //    force: used to force the scale operation in case of blocked deployment
 func (r *marathonClient) ScaleApplicationInstances(name string, instances int, force bool) (*DeploymentID, error) {
-	log.Printf("scaling the application: %s, instance: %d", name, instances)
 	changes := new(Application)
 	changes.ID = validateID(name)
 	changes.Instances = instances
@@ -514,13 +511,9 @@ func (r *marathonClient) ScaleApplicationInstances(name string, instances int, f
 // 		application:		the structure holding the application configuration
 func (r *marathonClient) UpdateApplication(application *Application) (*DeploymentID, error) {
 	result := new(DeploymentID)
-	log.Printf("updating application: %v", application)
-
 	uri := fmt.Sprintf("%s/%s", marathonAPIApps, trimRootPath(application.ID))
-
 	if err := r.apiPut(uri, &application, result); err != nil {
 		return nil, err
 	}
-
 	return result, nil
 }
--- a/application.go
+++ b/application.go
@@ -19,10 +19,9 @@ package marathon
 import (
 	"errors"
 	"fmt"
+	"log"
 	"net/url"
 	"time"
-
-	"github.com/golang/glog"
 )
 
 var (
@@ -480,7 +479,7 @@ func (r *marathonClient) DeleteApplication(name string) (*DeploymentID, error) {
 // RestartApplication performs a rolling restart of marathon application
 // 		name: 		the id used to identify the application
 func (r *marathonClient) RestartApplication(name string, force bool) (*DeploymentID, error) {
-	glog.V(debugLevel).Infof("restarting the application: %s, force: %s", name, force)
+	log.Printf("restarting the application: %s, force: %v", name, force)
 	deployment := new(DeploymentID)
 	var options struct {
 		Force bool `json:"force"`
@@ -498,7 +497,7 @@ func (r *marathonClient) RestartApplication(name string, force bool) (*Deploymen
 // 		instances:	the number of instances you wish to change to
 //    force: used to force the scale operation in case of blocked deployment
 func (r *marathonClient) ScaleApplicationInstances(name string, instances int, force bool) (*DeploymentID, error) {
-	glog.V(debugLevel).Infof("scaling the application: %s, instance: %d", name, instances)
+	log.Printf("scaling the application: %s, instance: %d", name, instances)
 	changes := new(Application)
 	changes.ID = validateID(name)
 	changes.Instances = instances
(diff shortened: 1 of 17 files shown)


client.go
```diff
@@ -205,151 +201,97 @@ func (r *marathonClient) Ping() (bool, error) {
...
+		return nil
 
-		// Add any basic auth and the content headers
-		if r.config.HTTPBasicAuthUser != "" {
-			request.SetBasicAuth(r.config.HTTPBasicAuthUser, r.config.HTTPBasicPassword)
-		}
-		request.Header.Add("Content-Type", "application/json")
-		request.Header.Add("Accept", "application/json")
+	} else if response.StatusCode == 404 {
+		return ErrDoesNotExist
```
timoreimann: The switch-case syntax was more concise and thus more readable. Is there any particular reason you had to replace it?

timoreimann:
A general remark: You introduced many smallish, unrelated changes into this single PR, which makes the review more challenging. Please consider splitting up separate concerns in future changes.

application.go
```diff
@@ -480,7 +478,6 @@ func (r *marathonClient) DeleteApplication(name string) (*DeploymentID, error) {
 // RestartApplication performs a rolling restart of marathon application
 // 		name: 		the id used to identify the application
 func (r *marathonClient) RestartApplication(name string, force bool) (*DeploymentID, error) {
-	glog.V(debugLevel).Infof("restarting the application: %s, force: %s", name, force)
```
timoreimann: Why not convert this and the following removals in this file into `log.*` statements?

client.go
mattes: Agreed. Brought it back.

application.go
mattes: Right now we log `RestartApplication`, `ScaleApplicationInstances` and `UpdateApplication` while there is no log for `DeleteApplication`, `CreateApplication`, `SetApplicationVersion` and all the other "helper" funcs in `applications.go`. 

I'd rather have no logs (and let the user decide what to log) than inconsistent logs. What happens is, people start relying on logs, and then get confused if one expected log is missing. Like `DeleteApplication` for example. (People in this context could be people not familiar with the code, like devops)

I also think that logs here would just generate noise. The func's input and output variables hold all state and the user is always able to receive that state without missing information. This is different for `httpRequest` for example where we log raw http response, which is not returned to the user.

mattes:
@timoreimann Yes, sorry for the big PR.

application.go
timoreimann: The inconsistency problem could be countered by adding more logs. ☺ I agree with you though that there's no real need to log here.

const.go
```diff
@@ -42,3 +44,7 @@ const (
 	// EventsTransportSSE activates stream events transport
 	EventsTransportSSE
 )
+
+func init() {
+	log.SetFlags(log.Ltime | log.Ldate | log.Lshortfile)
+}
```
eicca: Does it mean, that if a user will set different flags they will be overwritten? Or if another library does similar thing (setting flags in `init`) - the behaviour will be unexpected?
I think I would set in `init` only the values which belong to `go-marathon`.

eicca:
If the user has some custom logging, for example https://github.com/Sirupsen/logrus, would it be possible to plug it in into `go-marathon`?

As I understood, glog implements a special `CopyStandardLogTo`, so the question is about what to do if the library (or someone's implementation) doesn't support it? Suggest to implement the same https://github.com/golang/glog/blob/master/glog.go#L914?
Or we still can consider injecting the logger. As I said before, I don't know what's the best practice here.

examples/events_sse_transport/main.go
```diff
@@ -18,11 +18,9 @@ package main
 
 import (
 	"flag"
-	"time"
-
 	marathon "github.com/gambol99/go-marathon"
-
-	"github.com/golang/glog"
+	"log"
+	"time"
```
eicca: Can you please re-order imports everywhere according to https://github.com/golang/go/wiki/CodeReviewComments#imports

client.go
```diff
@@ -205,151 +201,99 @@ func (r *marathonClient) Ping() (bool, error) {
...
-	return status, "", fmt.Errorf("%s", errorMessage)
-}
-
-func (r *marathonClient) httpRequest(method, uri, body string) (int, string, *http.Response, error) {
-	var content string
-	var response *http.Response
+	if len(jsonBody) > 0 {
+		log.Printf("apiCall(): %v %v %s returned %v %s\n", request.Method, request.URL.String(), jsonBody, response.Status, oneLogLine(respBody))
+	} else {
+		log.Printf("apiCall(): %v %v returned %v %s\n", request.Method, request.URL.String(), response.Status, oneLogLine(respBody))
```
eicca: Another question: how to make it silent now? Before it was possible with setting debug level, now it will log every time.
IMO in this case we should either remove debugging statements or put the debugging mode into the config.

const.go
mattes: good point. removed this.

mattes:
@eicca Regarding your question about other logging libs like logrus. All they would need to do is to use: https://golang.org/pkg/log/#SetOutput (which is what glog does internally as well)

examples/events_sse_transport/main.go
mattes: Oh wow. That's cool. I only used gofmt so far. Added goimports to the stack now.

client.go
mattes: Hmm. You're right. I must have been selfish. :-) As soon as you don't use Glog, it will just appear on stderr. I think there are two options to solve this:

1. let the user handle this. `log.SetOutput(io.Discard)`. Problem here is that this will disable logging globally.
2. Add a `*Logger` field to the client struct and use this new logger for logging. I don't think that this approach will work with glog out of the box then though.

eicca: 1. Asking user to disable logging globally to turn off debug would be a bit uncomfortable.
2. `I don't think that this approach will work with glog out of the box then though. ` - I think it might be a good idea. By the default it'll work as it is now. If user wants to disable the log he will provide logger with the out set to `ioutil.Discard`. Maybe I miss something here?

eicca:
> Please consider splitting up separate concerns in future changes.

I second this, it's hard to review and discuss different issues at once.

In general it looks good to me and it's a good direction for the logging problem. We just need to generalize it a bit in case user doesn't use `glog` (for example me :) )

mattes:
> We just need to generalize it a bit in case user doesn't use glog

Will look into this asap.

mattes:
Added a `*log.logger` field to `marathonClient`. I had to remove `go 1.4` from the tests, as `log` has no `SetOutput` func in `go 1.4`.

eicca:
> I had to remove go 1.4 from the tests, as log has no SetOutput func in go 1.4. 

I'm not sure if we can drop 1.4 now. @gambol99 what do you think?

mattes:
Like the config idea. @eicca Will update it.

timoreimann:
Would love to see this making it out of the door. @gambol99, any chance you can have a look at it? Thanks.

gambol99:
@timoreimann apologizes for the delay ... became a dad last weekend, so things have been hectic to say the least ... i'll have a look now ..

timoreimann:
Congratulations @gambol99! Hope you and your family are doing fine. :baby:

No worries about the delay. There are things in life which clearly have priority.

(As a father of two, let me tell you that the best way to deal with sleep deprivation is to stop trying to get rest and spend the nights coding instead. ;-) ) 

Thanks for the update.

mattes:
@gambol99 Congrats!! :-)

gambol99:
thank you kindly :-) ... 

LGTM ..

PR #109, opened 2015-12-02
Refactor HasApplication to return only error.

eicca:
This PR refactors `HasApplication` to return only error instead of `(bool, err)`. It was inspired by how go standard library deals in similar situations: http://stackoverflow.com/a/12518877/1047843

Now after rebasing with latest master I realised that `HasApplication` doesn't do a lot and it might be easier to just call `Application` instead.

What do you think @gambol99 @timoreimann @mattes ?

### DIFF
--- a/application.go
+++ b/application.go
@@ -351,13 +351,6 @@ func (r *marathonClient) Application(name string) (*Application, error) {
 // If no health checks exist, we simply return true
 // 		name: 		the id used to identify the application
 func (r *marathonClient) ApplicationOK(name string) (bool, error) {
-	// step: check the application even exists
-	if found, err := r.HasApplication(name); err != nil {
-		return false, err
-	} else if !found {
-		return false, ErrDoesNotExist
-	}
-
 	// step: get the application
 	application, err := r.Application(name)
 	if err != nil {
@@ -431,36 +424,18 @@ func (r *marathonClient) WaitOnApplication(name string, timeout time.Duration) e
 			flick.SwitchOn()
 		}()
 		for !flick.IsSwitched() {
-			if found, err := r.HasApplication(name); err != nil {
+			app, err := r.Application(name)
+			if err != nil && err != ErrDoesNotExist {
 				continue
-			} else if found {
-				if app, err := r.Application(name); err == nil && app.AllTaskRunning() {
-					return nil
-				}
+			}
+			if err == nil && app.AllTaskRunning() {
+				return nil
 			}
 			time.Sleep(time.Duration(500) * time.Millisecond)
 		}
 		return nil
 	})
 	return err
-}
-
-// HasApplication checks to see if the application exists in marathon
-// 		name: 		the id used to identify the application
-func (r *marathonClient) HasApplication(name string) (bool, error) {
-	if name == "" {
-		return false, ErrInvalidArgument
-	}
-
-	if _, err := r.Application(name); err != nil {
-		if err == ErrDoesNotExist {
-			return false, nil
-		}
-
-		return false, err
-	}
-
-	return true, nil
 }
 
 // DeleteApplication deletes an application from marathon
--- a/application_test.go
+++ b/application_test.go
@@ -263,7 +263,7 @@ func TestApplicationOK(t *testing.T) {
 	assert.False(t, ok)
 }
 
-func TestListApplication(t *testing.T) {
+func TestApplication(t *testing.T) {
 	endpoint := newFakeMarathonEndpoint(t, nil)
 	defer endpoint.Close()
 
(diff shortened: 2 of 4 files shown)


mattes:
Makes sense. LGTM

timoreimann:
I'm also in favor of removing `HasApplication`: Ignoring the first parameter of `Application` is easy, and the input validation should occur with all functions taking a name parameter anyway. Plus, it'll simplify the code a bit.

I'd just add an example to the README file explaining how `Application` can be used to verify the existence of an app (by checking for the particular error).

application.go
```diff
@@ -431,12 +429,14 @@ func (r *marathonClient) WaitOnApplication(name string, timeout time.Duration) e
 			flick.SwitchOn()
 		}()
 		for !flick.IsSwitched() {
-			if found, err := r.HasApplication(name); err != nil {
-				continue
-			} else if found {
-				if app, err := r.Application(name); err == nil && app.AllTaskRunning() {
-					return nil
+			if err := r.HasApplication(name); err != nil {
```
timoreimann: I think `HasApplication` isn't really needed here: You can call `Application` and return values depending on the error received.

This is even more true if we agree on removing `HasApplication` in its entirety.

application.go
```diff
@@ -431,12 +429,14 @@ func (r *marathonClient) WaitOnApplication(name string, timeout time.Duration) e
...
+				if err != ErrDoesNotExist {
+					return err
```
timoreimann: The original code is a bit hard to understand, but I think you may have inadvertently deviated from the existing logic along your refactoring:

- When the application does not exist in the old implementation (`found` == `false` and `err` == `nil`), we sleep 500 ms. In the new implementation where _not found_ is exposed via the `ErrDoesNotExist` error, we just continue.
- When `HasApplication` returns an error in the old implementation (except for `ErrDoesNotExist` because that's mapped to a `false` result), we continue. In the new implementation, we return the error to the caller for anything other than `ErrDoesNotExist`.

eicca: Thanks for checking, looks like I completely misunderstood this part. I'll fix it! :)

application.go
eicca: I agree, I'm going to get rid of the `HasApllication` then.

application.go
timoreimann: You may want to check the tests for the method too. If I am right, then some test should have broken. If I misunderstood something, however, the tests should prove you right. So it's probably a good idea to double check our test coverage in this regards.

application.go
gambol99: @eicca makes sense ... since you can rely on the custom error from Application() ..

application.go
eicca: Unfortunately it's not tested. This method requires more sophisticated behaviour for test than we can provide with endpoint mocks. I'll try to look into it soon.

eicca:
@timoreimann can you have a second look please?

application_test.go
```diff
@@ -277,23 +277,17 @@ func TestListApplication(t *testing.T) {
...
+	_, err := endpoint.Client.Application("no_such_app")
+	assert.Equal(t, ErrDoesNotExist, err)
 
 	config := NewDefaultConfig()
 	config.URL = "http://non-existing-marathon-host.local:5555"
 	endpoint = newFakeMarathonEndpoint(t, &config)
 
-	found, err = endpoint.Client.HasApplication(fakeAppName)
+	_, err = endpoint.Client.Application(fakeAppName)
 	assert.Error(t, err)
```
timoreimann: I'd also assert here that `err` does not equal `ErrDoesNotExist` in order to guard against the case that we might have created another case for a non-existing application unwittingly, either now or in the future.

application_test.go
```diff
@@ -277,23 +277,17 @@ func TestListApplication(t *testing.T) {
 	assert.Equal(t, len(application.Tasks), 2)
 }
 
-func TestHasApplication(t *testing.T) {
+func TestApplicationDoesntExists(t *testing.T) {
```
timoreimann: I think the name is (partially) misleading because the second test in this function does not relate to a non-existing application (but rather a generic error case).

You could rename it to something like  `TestApplicationFailing` or consider merging the test with `TestListApplication`. The latter would be somewhat more consistent with the other tests where a single function covers both erroneous and non-erroneous outcomes.
(Side note: I think `TestListApplication` should be called `TestApplication` because `Application` is the interface method being tested. You may want to consider this if you decide to merge.)

eicca:
@timoreimann makes sense, done.

timoreimann:
LGTM.

I also just created https://github.com/gambol99/go-marathon/issues/110 so that we don't forget about increasing the test coverage adequately.

Issue #113, opened 2015-12-10
Application example does not remove app


jglobant:
Executing the applications-example a second time it gives me:
Failed, error: conflicting resource
I will commit a fix soon

jglobant:
replace this
```
	if _, err := client.Application(applicationName); err != nil {
		deployID, err := client.DeleteApplication(applicationName)
		assert(err)
		waitOnDeployment(client, deployID)
	}
```
with this
```
    if deployID, err := client.DeleteApplication(applicationName); err == nil {
        waitOnDeployment(client, deployID)
    }
```

jglobant:
close issue because of PR

PR #114, opened 2015-12-10
delete app before adding a new one

jglobant:
after showing all apps, the example will delete /my/product to create the same immediatly.
The bug was that the example results in "Failed, error: conflicting resource" because it tried to create the same app again which was already created the last run.

fix for issue #113


eicca:
Thanks for reporting this.

`_, err := client.Application(applicationName)` checks if the application exists or not. The problem is here, that it should check for `err == nil` instead of `err != nil`. I'd keep this check in order to show how to use this existence checking.

Running this example I've met a few other problems. We need to wait after every deployment:
```diff
--- a/examples/applications/main.go
+++ b/examples/applications/main.go
@@ -66,7 +66,7 @@ func main() {
 
        applicationName := "/my/product"
 
-       if _, err := client.Application(applicationName); err != nil {
+       if _, err := client.Application(applicationName); err == nil {
                deployID, err := client.DeleteApplication(applicationName)
                assert(err)
                waitOnDeployment(client, deployID)
@@ -83,6 +83,7 @@ func main() {
        application.Container.Docker.Container("quay.io/gambol99/apache-php:latest").Expose(80).Expose(443)
        _, err = client.CreateApplication(application)
        assert(err)
+       client.WaitOnApplication(application.ID, 30*time.Second)
 
        log.Printf("Scaling the application to 4 instances")
        deployID, err := client.ScaleApplicationInstances(application.ID, 4, false)
@@ -92,8 +93,9 @@ func main() {
 
        log.Printf("Deleting the application: %s", applicationName)
        deployID, err = client.DeleteApplication(application.ID)
+
        assert(err)
-       time.Sleep(time.Duration(10) * time.Second)
+       waitOnDeployment(client, deployID)
        log.Printf("Successfully deleted the application")
 
        log.Printf("Starting the application again")
```

If these changes make sense for you, could you please update and rebase PR?
Thanks.

eicca:
I opened another PR: https://github.com/gambol99/go-marathon/pull/120
Thanks again for reporting.

PR #115, opened 2015-12-11
Rename testing utils to avoid adding testing dependency

xperimental:
The file `testing_utils.go` only contains functions used by the testing code and should not be part of the normal compilation to avoid spilling the contents of the testing package into every binary compiled with this library.


eicca:
Makes sense.

What do you think about moving this code into another package called `testutils` and then using this package from tests? In this case:
- `*_test` files only contain tests
- other users can re-use `testutils` in order to test their code with `go-marathon`

xperimental:
Currently there are no public symbols in the file, so putting it into another package would require some refactoring. There might be some benefit for testing other code, but personally I think fixing the bug first is a better idea.

eicca:
LGTM

PR #116, opened 2015-12-13
Add docker-compose for examples.

eicca:
I added a docker-compose I'm using to test examples. Would it be helpful for others as well?


gambol99:
:+1: LGTM

Issue #112, opened 2015-12-10
kill Task does not work when scale=true is used. 


lokeshverizon:
KillTask(taskID string, scale bool) (*Task, error)

does not work properly with scale=true. (scale=true) does not take affect. Based on the debugging I found: rest API which is passed is: 
 http://10.1.42.1:19090/v2/apps/appid/tasks/taskid  {"bool":true} 

it should be 
http://162.47.116.88:19090/v2/apps//appid/tasks/taskid?scale=true

As marathon GUI uses kill&scale on task above URL.

eicca:
I can confirm this, marathon expects parameters to be in the url.

I started fixing the bug here: https://github.com/gambol99/go-marathon/compare/master...eicca:refactor-tasks

I'm going to prepare a PR soon.

PR #117, opened 2015-12-15
Refactor tasks

eicca:
I tried to go through `tasks.go` and refactor/fix some code:
- fixes https://github.com/gambol99/go-marathon/issues/112
- fixes return value for `KillTask`
- introduced options for requests, removed url string manipulation
- added missing tests
- added tasks example
- refactored here and there

> introduced options for requests

I remember we discussed it once, however I went back to this idea after trying to understand what `AllTasks(v url.Values)` does. In order to use this method, one needs to:
- check go `url.Values` documentation
- check marathon rest api documentation (which the library tries to abstract)
- combine them together

I don't think that it would be a problem to update the library if marathon will introduce new parameters. Moreover, it happens quite rarely.

If you don't agree with the options struct, I'm happy to move it back to `url.Values`.

### DIFF
--- a/application.go
+++ b/application.go
@@ -192,12 +192,9 @@ func (r *Application) AddLabel(name, value string) *Application {
 	return r
 }
 
-// HasHealthChecks is more of a helper method, used to check if an application has healtchecks
+// HasHealthChecks is a helper method, used to check if an application has healtchecks
 func (r *Application) HasHealthChecks() bool {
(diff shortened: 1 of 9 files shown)


gambol99:
Admittedly I prefer the url.Value option because it's a free of all :-), but the above approach is probably the right way of doing it.. :+1: LGTM

eicca:
Thanks for reviewing!

I'll try to make it consistent across the library.

PR #119, opened 2015-12-18
Fixed minor issue in ScaleApplicationInstances example

h-marvin:
- ScaleApplicationInstances example no uses the value that the readme states it will


timoreimann:
LGTM.

eicca:
Thanks!

PR #120, opened 2015-12-22
Fix application example.

eicca:
Fix for https://github.com/gambol99/go-marathon/pull/114


eicca:
I will merge this fix, since the example in the current state might be misleading.

PR #121, opened 2016-01-13
Check for empty Marathon hosts; include tests for newCluster.

### LINKED ISSUE #104: newCluster() broken code
https://github.com/gambol99/go-marathon/blob/master/cluster.go#L90

`newCluster()` should first split by commas and then do the `url.Parse()`. I'm surprised that url.Parse isn't failing anyway. The marathon variables holds:

```
&url.URL{Scheme:"http", Opaque:"", User:(*url.Userinfo)(nil), Host:"10.241.1.71:8080,10.241.1.72:8080,10.241.1.73:8080", Path:"", RawPath:"", RawQuery:"", Fragment:""}
```

... and `Host` doesn't look like a valid host to me.

timoreimann:
The newCluster was missing test coverage and would not mark empty hosts as an error. This change fixes both.

Closes #104

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -102,6 +102,10 @@ func newCluster(marathonURL string) (Cluster, error) {
 	/* step: create a link list of the hosts */
 	var previous *marathonNode
 	for index, host := range strings.SplitN(marathon.Host, ",", -1) {
+		if len(host) == 0 {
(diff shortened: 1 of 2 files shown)


gambol99:
LGTM

PR #124, opened 2016-01-21
Fix scaling application to zero instances

xperimental:
I noticed that the `ScaleApplicationInstances` as well as the `UpdateApplication` are not able to set the application instances to zero. This is due to the `omitempty` annotation on the `Instances` field in the `Application` struct.

I did not want to remove the omitempty from the Application struct itself, as you always would have to set the correct instance count when doing `UpdateApplication` to avoid suspending your application. Instead I just changed the `ScaleApplicationInstances` method to use an anonymous struct which does not have that annotation and also only contains the necessary fields for scaling an application. That way you have to explicitly use `ScaleApplicationInstances` if you want to suspend your application.

### DIFF
--- a/application.go
+++ b/application.go
@@ -464,9 +464,13 @@ func (r *marathonClient) RestartApplication(name string, force bool) (*Deploymen
 // 		instances:	the number of instances you wish to change to
 //    force: used to force the scale operation in case of blocked deployment
 func (r *marathonClient) ScaleApplicationInstances(name string, instances int, force bool) (*DeploymentID, error) {
-	changes := new(Application)
-	changes.ID = validateID(name)
-	changes.Instances = instances
+	changes := struct {
+		ID        string `json:"id"`
+		Instances int    `json:"instances"`
+	}{
+		ID:        validateID(name),
+		Instances: instances,
+	}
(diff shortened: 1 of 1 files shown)


kesselborn:
I am running into the same problem: I would like to create an app in marathon and not immediately scale up one instance -- therefore I tried setting the instance count to 0 ... which fails due to the same problem.
I actually think that it would make sense to actually remove the `omitempty` annotation from instances. If somebody wants to keep the instance count, it is easy enough to query the app before updating.

kesselborn:
mmm ... ok ... perhaps it could mean that updates actually scale down to 0 if somebody only provides the diff

timoreimann:
@gambol99 @eicca Chance looks good to me. Do you think we can consider merging this PR?

eicca:
I apologize for a lack of activity.

The change looks good to me as a bugfix.

In general it could be more correct to:
- get rid of `ScaleApplicationInstances` and use update directly
- remove `omitempty` where it breaks the behaviour

It would be also more consistent with marathon api (https://mesosphere.github.io/marathon/docs/rest-api.html), since it doesn't have an action for scaling.

gambol99:
LGTM

PR #125, opened 2016-01-26
skip tls verification

gambol99:
- adding the skip tls verify option to the client

### DIFF
--- a/client.go
+++ b/client.go
@@ -165,22 +165,23 @@ type marathonClient struct {
 // NewClient creates a new marathon client
 //		config:			the configuration to use
 func NewClient(config Config) (Marathon, error) {
-	// step: we parse the url and build a cluster
-	cluster, err := newCluster(config.URL)
+	// step: if no http client, set to default
+	if config.HTTPClient == nil {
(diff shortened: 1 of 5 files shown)


client.go
```diff
@@ -183,7 +185,20 @@ func NewClient(config Config) (Marathon, error) {
...
+			Proxy: http.ProxyFromEnvironment,
+			Dial: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).Dial,
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: config.SkipTLSVerify,
+			},
+			TLSHandshakeTimeout: 10 * time.Second,
+		},
```
eicca: Could it be easier just to have an option to specify your own `httpClient` in config and fall-back to default. In this case user can provide all necessary options (like self-signed certificates) and we just need to provide an example how to do it.

gambol99:
@eicca i agree! ... Will change to allow override the default http client

eicca:
LGTM!

PR #127, opened 2016-02-09
Fix issue with setting empty values

kesselborn:
Make properties pointers if sending an empty value (int -> 0, string ->
"") to marathon has a semantic meaning. In order to be able to distinguish between
null-values and unset-values (see go-github blog post discussing the
same problem for go-github [0]) -- it's actually the same problem described
in gambol99/go-marathon#124 but for all^W most properties.

Instead of providing convenience functions such as
`github.String("string")` that return a pointer to a string, I added
methods that use the builder pattern'ish approach that quite a few
properties had before as well, i.e. construction looks something like:

    application := marathon.NewDockerApplication().
      Name("myapp").
      Command("/hello-world-server").
      AddEnv("PORT", "8080").
      ...
      Count(0)

    application.Container.Docker.Container("alpine-hello-world-server").
      Bridged().
      ExposePort(8080, 0, 0, "tcp").
      SetPrivileged(false).
      SetForcePullImage(false).
      Parameter("ulimit", fmt.Sprintf("nofile=%d", concern.Limits.FileMax))

There are specialized methods called for example `EmptyArgs()` to explicitly set
an array or hash to be empty (vs. omiting it, which would mean it keeps the current
value).

For consistency's sake, I renamed `Arg` -> 'AddArgs' and `Parameter` ->
`AddParameter` as this change is a breaking change anyways.

[0] https://willnorris.com/2014/05/go-rest-apis-and-pointers

### DIFF
--- a/application.go
+++ b/application.go
@@ -17,6 +17,7 @@ limitations under the License.
 package marathon
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/url"
@@ -36,23 +37,23 @@ type Applications struct {
 // Application is the definition for an application in marathon
 type Application struct {
 	ID                    string              `json:"id,omitempty"`
-	Cmd                   string              `json:"cmd,omitempty"`
-	Args                  []string            `json:"args"`
-	Constraints           [][]string          `json:"constraints"`
+	Cmd                   *string             `json:"cmd,omitempty"`
+	Args                  *[]string           `json:"args,omitempty"`
+	Constraints           *[][]string         `json:"constraints,omitempty"`
 	Container             *Container          `json:"container,omitempty"`
 	CPUs                  float64             `json:"cpus,omitempty"`
-	Disk                  float64             `json:"disk,omitempty"`
-	Env                   map[string]string   `json:"env"`
-	Executor              string              `json:"executor,omitempty"`
-	HealthChecks          []*HealthCheck      `json:"healthChecks"`
-	Instances             int                 `json:"instances,omitempty"`
-	Mem                   float64             `json:"mem,omitempty"`
+	Disk                  *float64            `json:"disk,omitempty"`
+	Env                   *map[string]string  `json:"env,omitempty"`
+	Executor              *string             `json:"executor,omitempty"`
+	HealthChecks          *[]HealthCheck      `json:"healthChecks,omitempty"`
+	Instances             *int                `json:"instances,omitempty"`
+	Mem                   *float64            `json:"mem,omitempty"`
 	Tasks                 []*Task             `json:"tasks,omitempty"`
 	Ports                 []int               `json:"ports"`
-	RequirePorts          bool                `json:"requirePorts,omitempty"`
-	BackoffSeconds        float64             `json:"backoffSeconds,omitempty"`
-	BackoffFactor         float64             `json:"backoffFactor,omitempty"`
-	MaxLaunchDelaySeconds float64             `json:"maxLaunchDelaySeconds,omitempty"`
+	RequirePorts          *bool               `json:"requirePorts,omitempty"`
+	BackoffSeconds        *float64            `json:"backoffSeconds,omitempty"`
+	BackoffFactor         *float64            `json:"backoffFactor,omitempty"`
+	MaxLaunchDelaySeconds *float64            `json:"maxLaunchDelaySeconds,omitempty"`
 	Deployments           []map[string]string `json:"deployments,omitempty"`
 	Dependencies          []string            `json:"dependencies"`
 	TasksRunning          int                 `json:"tasksRunning,omitempty"`
@@ -61,10 +62,10 @@ type Application struct {
 	TasksUnhealthy        int                 `json:"tasksUnhealthy,omitempty"`
 	User                  string              `json:"user,omitempty"`
 	UpgradeStrategy       *UpgradeStrategy    `json:"upgradeStrategy,omitempty"`
-	Uris                  []string            `json:"uris"`
+	Uris                  *[]string           `json:"uris"`
 	Version               string              `json:"version,omitempty"`
 	VersionInfo           *VersionInfo        `json:"versionInfo,omitempty"`
-	Labels                map[string]string   `json:"labels,omitempty"`
+	Labels                *map[string]string  `json:"labels,omitempty"`
 	AcceptedResourceRoles []string            `json:"acceptedResourceRoles,omitempty"`
 	LastTaskFailure       *LastTaskFailure    `json:"lastTaskFailure,omitempty"`
 	Fetch                 []Fetch             `json:"fetch"`
@@ -101,9 +102,15 @@ func NewDockerApplication() *Application {
 	return application
 }
 
-// Name set the name of the application i.e. the identifier for this application
+// Name sets the name / ID of the application i.e. the identifier for this application
 func (r *Application) Name(id string) *Application {
 	r.ID = validateID(id)
+	return r
+}
+
+// Command sets the cmd of the application
+func (r *Application) Command(cmd string) *Application {
+	r.Cmd = &cmd
 	return r
 }
 
@@ -118,20 +125,20 @@ func (r *Application) CPU(cpu float64) *Application {
 // application I don't believe is relevant
 //		disk:	the disk space in MB
 func (r *Application) Storage(disk float64) *Application {
-	r.Disk = disk
+	r.Disk = &disk
 	return r
 }
 
 // AllTaskRunning checks to see if all the application tasks are running, i.e. the instances is equal
 // to the number of running tasks
 func (r *Application) AllTaskRunning() bool {
-	if r.Instances == 0 {
+	if r.Instances == nil || *r.Instances == 0 {
 		return true
 	}
 	if r.Tasks == nil {
 		return false
 	}
-	if r.TasksRunning == r.Instances {
+	if r.TasksRunning == *r.Instances {
 		return true
 	}
 	return false
@@ -153,7 +160,7 @@ func (r *Application) DependsOn(names ...string) *Application {
 // Memory sets he amount of memory the application can consume per instance
 //		memory:	the amount of MB to assign
 func (r *Application) Memory(memory float64) *Application {
-	r.Mem = memory
+	r.Mem = &memory
 
 	return r
 }
@@ -161,18 +168,74 @@ func (r *Application) Memory(memory float64) *Application {
 // Count sets the number of instances of the application to run
 //		count:	the number of instances to run
 func (r *Application) Count(count int) *Application {
-	r.Instances = count
+	r.Instances = &count
 
 	return r
 }
 
-// Arg adds one or more arguments to the applications
+// AddArgs adds one or more arguments to the applications
 //		arguments:	the argument(s) you are adding
-func (r *Application) Arg(arguments ...string) *Application {
+func (r *Application) AddArgs(arguments ...string) *Application {
 	if r.Args == nil {
-		r.Args = make([]string, 0)
+		r.EmptyArgs()
 	}
-	r.Args = append(r.Args, arguments...)
+
+	args := *r.Args
+	args = append(args, arguments...)
+	r.Args = &args
+
+	return r
+}
+
+// EmptyArgs explicitly empties arguments -- use this if you need to empty
+// arguments of an application that already has arguments set (setting args to nil will
+// keep the current value)
+func (r *Application) EmptyArgs() *Application {
+	r.Args = &[]string{}
+
+	return r
+}
+
+// AddConstraint adds a new constraint
+//		arguments:	the constraint definition, one argument per array element
+func (r *Application) AddConstraint(arguments ...string) *Application {
+	if r.Constraints == nil {
+		r.EmptyConstraints()
+	}
+
+	constraints := *r.Constraints
+	constraints = append(constraints, arguments)
+	r.Constraints = &constraints
+
+	return r
+}
+
+// EmptyConstraints explicitly empties constraints -- use this if you need to empty
+// constraints of an application that already has constraints set (setting constraints to nil will
+// keep the current value)
+func (r *Application) EmptyConstraints() *Application {
+	r.Constraints = &[][]string{}
+
+	return r
+}
+
+// AddLabel adds a label to the application
+//		name:	the name of the label
+//		value: value for this label
+func (r *Application) AddLabel(name, value string) *Application {
+	if r.Labels == nil {
+		r.EmptyLabel()
+	}
+	(*r.Labels)[name] = value
+
+	return r
+}
+
+// EmptyLabel explicitly empties the labels -- use this if you need to empty
+// the labels of an application that already has labels set (setting labels to nil will
+// keep the current value)
+func (r *Application) EmptyLabel() *Application {
+	r.Labels = &map[string]string{}
 
 	return r
 }
@@ -182,28 +245,55 @@ func (r *Application) Arg(arguments ...string) *Application {
 //		value:	go figure, the value associated to the above
 func (r *Application) AddEnv(name, value string) *Application {
 	if r.Env == nil {
-		r.Env = make(map[string]string, 0)
+		r.EmptyEnv()
 	}
-	r.Env[name] = value
+	(*r.Env)[name] = value
 
 	return r
 }
 
-// AddLabel adds a label to the application
-//		name:	the name of the label
-//		value:	go figure, the value associated to the above
-func (r *Application) AddLabel(name, value string) *Application {
-	if r.Labels == nil {
-		r.Labels = make(map[string]string, 0)
+// EmptyEnv explicitly empties the env -- use this if you need to empty
+// the environment of an application that already has an environment set (setting env to nil will
+// keep the current value)
+func (r *Application) EmptyEnv() *Application {
+	r.Env = &map[string]string{}
+
+	return r
+}
+
+// SetExecutor sets the executor
+func (r *Application) SetExecutor(executor string) *Application {
+	r.Executor = &executor
+
+	return r
+}
+
+// AddHealthCheck adds a health check
+// 	healtchecks the health check that should be added
+func (r *Application) AddHealthCheck(healthCheck HealthCheck) *Application {
+	if r.HealthChecks == nil {
+		r.EmptyHealthChecks()
 	}
-	r.Labels[name] = value
+
+	healthChecks := *r.HealthChecks
+	healthChecks = append(healthChecks, healthCheck)
+	r.HealthChecks = &healthChecks
+
+	return r
+}
+
+// EmptyHealthChecks explicitly empties healthChecks -- use this if you need to empty
+// healthChecks of an application that already has healthChecks set (setting healthChecks to nil will
+// keep the current value)
+func (r *Application) EmptyHealthChecks() *Application {
+	r.HealthChecks = &[]HealthCheck{}
 
 	return r
 }
 
 // HasHealthChecks is a helper method, used to check if an application has healtchecks
 func (r *Application) HasHealthChecks() bool {
-	return r.HealthChecks != nil && len(r.HealthChecks) > 0
+	return r.HealthChecks != nil && len(*r.HealthChecks) > 0
 }
 
 // DeploymentIDs retrieves the application deployments IDs
@@ -230,9 +320,6 @@ func (r *Application) DeploymentIDs() []*DeploymentID {
 //		port: 		the port the check should be checking
 // 		interval:	the interval in seconds the check should be performed
 func (r *Application) CheckHTTP(uri string, port, interval int) (*Application, error) {
-	if r.HealthChecks == nil {
-		r.HealthChecks = make([]*HealthCheck, 0)
-	}
 	if r.Container == nil || r.Container.Docker == nil {
 		return nil, ErrNoApplicationContainer
 	}
@@ -242,11 +329,11 @@ func (r *Application) CheckHTTP(uri string, port, interval int) (*Application, e
 		return nil, err
 	}
 	health := NewDefaultHealthCheck()
-	health.Path = uri
 	health.IntervalSeconds = interval
-	health.PortIndex = portIndex
+	*health.Path = uri
+	*health.PortIndex = portIndex
 	// step: add to the checks
-	r.HealthChecks = append(r.HealthChecks, health)
+	r.AddHealthCheck(*health)
 
 	return r, nil
 }
@@ -256,9 +343,6 @@ func (r *Application) CheckHTTP(uri string, port, interval int) (*Application, e
 //		port: 		the port the check should, err, check
 // 		interval:	the interval in seconds the check should be performed
 func (r *Application) CheckTCP(port, interval int) (*Application, error) {
-	if r.HealthChecks == nil {
-		r.HealthChecks = make([]*HealthCheck, 0)
-	}
 	if r.Container == nil || r.Container.Docker == nil {
 		return nil, ErrNoApplicationContainer
 	}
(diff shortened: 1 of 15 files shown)


timoreimann:
I'm generally in favor of the go-github approach to use pointers in order to distinguish field absence and zero values. I'm wondering though if we need it to the extent that it was implemented in this PR.

For instance, passing an empty `ID` should never be a meaningful operation. Same probably goes for `CPUs`, `Disk`, and `Mem`. Relying on zero values getting omitted in the resulting marshaled request for such values should always be a safe operation and do no harm. Limiting where we use pointers should simplify handling of the data and also make clear whether an zero value carries semantics relevant to the Marathon API or not.

WDYT?

kesselborn:
@timoreimann thanks for the quick comment.
I think you have a good point -- I think in the end I went for consistency for all read-write-able properties, but expressing whether zero values carry semantics through mixing pointer and non-pointer types kind of makes sense as well (it's probably not obvious on first sight and a little bit confusing why some properties are values and others are pointers when starting with the lib though).
For the app, I see the need to check the following properties regarding types:
```
Check whether null-values make sense:
	Container             
	CPUs                  
	Disk                  
	BackoffSeconds        
	BackoffFactor         
	MaxLaunchDelaySeconds 
	UpgradeStrategy       
	Version               

Read only?:
	TasksRunning          
	TasksStaged           
	TasksHealthy          
	TasksUnhealthy        
	User                  
	LastTaskFailure       
	VersionInfo           
```

timoreimann:
Agree with you that the mixed approach looks inconsistent on first sight. Apart from clarifying semantics, however, I think it will reduce implementation complexity and simplify client usage, so the benefits seem worth it from my perspective.

Unfortunately, the official Marathon REST documentation isn't perfectly clear on all of the open points. I think we have two possible ways to move forward from here:

1. invest time into figuring out which parameters have zero value semantics and pointerize them; or
1. pointerize the parameters which we know for sure have zero value semantics; keep everything else untouched and only make changes on a as-needed basis.

While the first option would be ideal, I personally would be perfectly fine with the second one. Up to you. :-)

timoreimann:
Oh, regarding the possibly read-only values: I'm actually wondering why these are part of the `Application` struct. They only seem to show up in responses, and thus could be part of a separate struct as far as I can see.

kesselborn:
@timoreimann sorry for the long delay: I know revisited this PR and made the changes as discussed:

Only make property pointer if zero-value makes sense …
Changing back the following properties to non-pointer-types:
- `Application.ID`: id is always required -- setting this to an empty string will fail
- `Application.CPUs`: the CPUs property needs to be a value > 0, i.e. 0 is forbidden
- in Application struct: BackoffSeconds, BackoffFactor and MaxLaunchDelaySeconds, Disk and Memory can all be set to zero-values
  although this does not really make sense. However: I think we should support whatever marathon
  accepts just in case zero values should have the meaning of: "don't know, you decide"
- in Application struct: TasksRunning, TasksStaged, TasksHealthy, TasksUnhealthy, User and Version are read only properties
  and thuse don't need to be pointerized
- Docker.Image and Docker.Network must not have zero value set, i.e. un-pointerize them
- Group.ID and Groups.ID must not be zero values, i.e. un-pointerize them

... note that this PR unfortunately includes https://github.com/gambol99/go-marathon/pull/132 as I did not want to create a different pull request + I did not our project to not compile anymore. Before merging this, I can clean this up.

application.go
```diff
@@ -36,23 +37,23 @@ type Applications struct {
 // Application is the definition for an application in marathon
 type Application struct {
 	ID                    string              `json:"id,omitempty"`
-	Cmd                   string              `json:"cmd,omitempty"`
-	Args                  []string            `json:"args"`
-	Constraints           [][]string          `json:"constraints"`
+	Cmd                   *string             `json:"cmd,omitempty"`
+	Args                  *[]string           `json:"args,omitempty"`
```
timoreimann: According to the documentation, either `Cmd` or `Args` must be provided. I haven't tested it but would assume that if you have `Cmd` set on an application and pass `Args` in a `PUT` request, Marathon will use `Args` now (and vice versa).

Have you possibly tested this?

docker.go
```diff
@@ -142,24 +193,37 @@ func (docker *Docker) ExposePort(containerPort, hostPort, servicePort int, proto
 //		value:		the value of the option
 func (docker *Docker) Parameter(key string, value string) *Docker {
```
timoreimann: Would be nice to add tests for the method while we touch it.

health.go
```diff
@@ -19,13 +19,13 @@ package marathon
 // HealthCheck is the definition for an application health check
 type HealthCheck struct {
 	Command                *Command `json:"command,omitempty"`
-	Protocol               string   `json:"protocol,omitempty"`
-	Path                   string   `json:"path,omitempty"`
-	GracePeriodSeconds     int      `json:"gracePeriodSeconds,omitempty"`
-	IntervalSeconds        int      `json:"intervalSeconds,omitempty"`
-	PortIndex              int      `json:"portIndex,omitempty"`
+	Protocol               string   `json:"protocol,"`
```
timoreimann: Removing `omitempty` means we cannot let the client rely on Marathon's default settings anymore. Therefore, I would pointerize the fields which have zero value semantics (namely, `GracePeriodSeconds` and `PortIndex`), provide convenience setters, and leave the rest untouched.

In case we decide to stick to the current approach: `Protocol` has a trailing comma after the tag specifiers.

utils.go
```diff
@@ -45,7 +45,8 @@ func (r *atomicSwitch) SwitchedOff() {
 
 func validateID(id string) string {
 	if !strings.HasPrefix(id, "/") {
-		return fmt.Sprintf("/%s", id)
+		correctedId := fmt.Sprintf("/%s", id)
+		return correctedId
```
timoreimann: Why the extra variable?

application.go
```diff
@@ -61,7 +62,7 @@ type Application struct {
 	TasksUnhealthy        int                 `json:"tasksUnhealthy,omitempty"`
 	User                  string              `json:"user,omitempty"`
 	UpgradeStrategy       *UpgradeStrategy    `json:"upgradeStrategy,omitempty"`
-	Uris                  []string            `json:"uris"`
+	Uris                  *[]string           `json:"uris"`
 	Version               string              `json:"version,omitempty"`
 	VersionInfo           *VersionInfo        `json:"versionInfo,omitempty"`
 	Labels                map[string]string   `json:"labels,omitempty"`
```
timoreimann: I think we need `Labels` to support zero values so that we can remove all labels in a `PUT` request.

application.go
```diff
@@ -101,12 +102,18 @@ func NewDockerApplication() *Application {
...
 
-// Name set the name of the application i.e. the identifier for this application
+// Name sets the name / ID of the application i.e. the identifier for this application
 func (r *Application) Name(id string) *Application {
 	r.ID = validateID(id)
 	return r
 }
 
+// Command sets the cmd of the application
+func (r *Application) Command(cmd string) *Application {
```
timoreimann: Method is missing test coverage.

application.go
```diff
@@ -118,20 +125,20 @@ func (r *Application) CPU(cpu float64) *Application {
...
 //		disk:	the disk space in MB
 func (r *Application) Storage(disk float64) *Application {
-	r.Disk = disk
+	r.Disk = &disk
 	return r
 }
 
 // AllTaskRunning checks to see if all the application tasks are running, i.e. the instances is equal
 // to the number of running tasks
 func (r *Application) AllTaskRunning() bool {
```
timoreimann: Would be nice to add tests for the method while we touch it.

application.go
```diff
@@ -161,18 +168,55 @@ func (r *Application) Memory(memory float64) *Application {
...
+
+	return r
+}
+
+// EmptyArgs explicitly empties arguments -- use this if you need to empty
+// arguments of an application that already has arguments set (setting args to nil will
+// keep the current value)
+func (r *Application) EmptyArgs() *Application {
+	empty := []string{}
+	r.Args = &empty
```
timoreimann: Can be simplified by inlining the address of the empty value. Specifically,

`r.Args = &[]string{}`

application.go
```diff
@@ -161,18 +168,55 @@ func (r *Application) Memory(memory float64) *Application {
...
+
+	return r
+}
+
+// AddConstraints adds a new constraint
```
timoreimann: Comment should start with `AddConstraint` (no plural).

application.go
```diff
@@ -161,18 +168,55 @@ func (r *Application) Memory(memory float64) *Application {
...
+
+	return r
+}
+
+// EmptyConstraints explicitly empties constraints -- use this if you need to empty
+// constraints of an application that already has constraints set (setting constraints to nil will
+// keep the current value)
+func (r *Application) EmptyConstraints() *Application {
+	empty := [][]string{}
+	r.Constraints = &empty
```
timoreimann: Can be simplified by inlining the address of the empty value.

application.go
```diff
@@ -182,9 +226,50 @@ func (r *Application) Arg(arguments ...string) *Application {
...
+
+	return r
+}
+
+// EmptyHealthChecks explicitly empties healthChecks -- use this if you need to empty
+// healthChecks of an application that already has healthChecks set (setting healthChecks to nil will
+// keep the current value)
+func (r *Application) EmptyHealthChecks() *Application {
+	healthChecks := []HealthCheck{}
+	r.HealthChecks = &healthChecks
```
timoreimann: Can be simplified by inlining the address of the empty value.

application.go
```diff
@@ -272,11 +351,45 @@ func (r *Application) CheckTCP(port, interval int) (*Application, error) {
...
 	// step: add to the checks
-	r.HealthChecks = append(r.HealthChecks, health)
+	r.AddHealthCheck(*health)
 
 	return r, nil
 }
 
+// AddUris adds one or more uris to the applications
+//		arguments:	the uri(s) you are adding
+func (r *Application) AddUris(newUris ...string) *Application {
```
timoreimann: Initialisms in Go are normally all-caps ([reference](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)), so I'd name it `[aA]ddURIs` in comments, methods, and variables. I know that's not consistent with other parts of go-marathon but believe we should start trying to be more in line with existing Go conventions.

application.go
```diff
@@ -272,11 +351,45 @@ func (r *Application) CheckTCP(port, interval int) (*Application, error) {
...
+	uris = append(uris, newUris...)
+	r.Uris = &uris
+
+	return r
+}
+
+// EmptyUris explicitly empties uris -- use this if you need to empty
+// uris of an application that already has uris set (setting uris to nil will
+// keep the current value)
+func (r *Application) EmptyUris() *Application {
```
timoreimann: `EmptyURIs` here too.

application.go
```diff
@@ -272,11 +351,45 @@ func (r *Application) CheckTCP(port, interval int) (*Application, error) {
...
+	empty := []string{}
+	r.Uris = &empty
```
timoreimann: Can be simplified by inlining the address of the empty value.

application.go
```diff
@@ -272,11 +351,45 @@ func (r *Application) CheckTCP(port, interval int) (*Application, error) {
...
+
+	return r
+}
+
+// String returns the json representation of this application
+func (r *Application) String() string {
```
timoreimann: Where do we use this?

If we do somewhere: I believe `String()` should be reserved for stringifying a Go type, so this one may be better called `JSON()`. It's also missing test coverage.

application.go
```diff
@@ -353,6 +466,19 @@ func (r *marathonClient) Application(name string) (*Application, error) {
 	return wrapper.Application, nil
 }
 
+// ApplicationConfiguration retrieves the application configuration from marathon
+// 		name: 		the id used to identify the application
+// 		version:  the version of the configuration you would like to receive
+func (r *marathonClient) ApplicationConfiguration(name, version string) (*Application, error) {
```
timoreimann: As you mentioned, this needs to be removed prior to merging.

application_test.go
```diff
@@ -65,19 +65,76 @@ func TestApplicationCPU(t *testing.T) {
...
+	assert.Equal(t, "-p", (*app.Args)[0])
+	assert.Equal(t, "option", (*app.Args)[1])
+	assert.Equal(t, "-v", (*app.Args)[2])
+
+	app.EmptyArgs()
+	assert.NotNil(t, app.Args)
+	assert.Equal(t, 0, len(*app.Args))
+}
+
+func ExampleApplication_AddConstraint() {
```
timoreimann: What's the purpose of this function? Seems like its functionality is already covered by `TestApplicationConstraints()`.

application_test.go
```diff
@@ -65,19 +65,76 @@ func TestApplicationCPU(t *testing.T) {
...
+	app.EmptyConstraints()
+	assert.NotNil(t, app.Constraints)
+	assert.Equal(t, 0, len(*app.Constraints))
 }
 
 func TestApplicationEnvs(t *testing.T) {
 	app := NewDockerApplication()
-	assert.Equal(t, 0, len(app.Env))
+	assert.Nil(t, app.Env)
 	app.AddEnv("hello", "world")
```
timoreimann: You could add another environment variable to verify that multiple variables are working.

application_test.go
```diff
@@ -213,6 +264,20 @@ func TestRestartApplication(t *testing.T) {
 	assert.Nil(t, id)
 }
 
+func TestApplicationUris(t *testing.T) {
```
timoreimann: Rename to `TestApplicationURIs()`.

docker.go
```diff
@@ -65,13 +65,28 @@ type Docker struct {
...
+
+	return container
+}
+
+// EmptyVolumes explicitly empties the volumes -- use this if you need to empty
+// volumes of an application that already has volumes set (setting volumes to nil will
+// keep the current value)
+func (container *Container) EmptyVolumes() *Container {
+	empty := []Volume{}
+	container.Volumes = &empty
```
timoreimann: Can be simplified by inlining the address of the empty value.

docker.go
```diff
@@ -79,16 +94,36 @@ func (container *Container) Volume(hostPath, containerPath, mode string) *Contai
...
-		Network:      "BRIDGE",
-		PortMappings: make([]*PortMapping, 0),
-		Parameters:   make([]*Parameters, 0),
-	}
-	container.Volumes = make([]*Volume, 0)
+	container.EmptyVolumes()
+	container.Docker = &Docker{}
+	container.Docker.Container("").
+		Bridged().
+		SetPrivileged(false).
```
timoreimann: Shouldn't we leave out all zero value-setting method calls from here onwards? I'm thinking of two use cases:

- Clients may be interested in letting Marathon pick default values.
- Clients may want to `PUT` an existing application configuration and only change individual fields.

docker.go
```diff
@@ -79,16 +94,36 @@ func (container *Container) Volume(hostPath, containerPath, mode string) *Contai
...
+		SetForcePullImage(false).
+		EmptyParameters().
+		EmptyPortMappings()
+
 	return container
 }
 
+// SetForcePullImage sets whether the docker image should always be force pulled before
+// starting an instance
+//		forcePull:			forceImagePull true / false (false by default)
```
timoreimann: If the `(false by default)` refers to Marathon's default, then I would leave it out as it could change over time. Clients should read up on the official REST documentation in order to learn about the default behavior.

docker.go
```diff
@@ -79,16 +94,36 @@ func (container *Container) Volume(hostPath, containerPath, mode string) *Contai
...
+func (docker *Docker) SetForcePullImage(forcePull bool) *Docker {
+	docker.ForcePullImage = &forcePull
+
+	return docker
+}
+
+// SetPrivileged sets whether the docker image should be started
+// with privilege turned on
+//		priv:			privilege true / false (false by default)
```
timoreimann: Again, I would leave out the default description for reasons explained above.

docker.go
```diff
@@ -98,7 +133,8 @@ func (docker *Docker) Container(image string) *Docker {
 
 // Bridged sets the networking mode to bridged
 func (docker *Docker) Bridged() *Docker {
-	docker.Network = "BRIDGE"
+	n := "BRIDGE"
+	docker.Network = n
```
timoreimann: No need for the extra variable `n`.

docker.go
```diff
@@ -127,13 +163,28 @@ func (docker *Docker) ExposeUDP(ports ...int) *Docker {
 //		protocol:						the protocol to use TCP, UDP
 func (docker *Docker) ExposePort(containerPort, hostPort, servicePort int, protocol string) *Docker {
 	if docker.PortMappings == nil {
-		docker.PortMappings = make([]*PortMapping, 0)
+		docker.EmptyPortMappings()
 	}
-	docker.PortMappings = append(docker.PortMappings, &PortMapping{
+
+	portMappings := *docker.PortMappings
+
```
timoreimann: nit-pick: Remove the line break.

docker.go
```diff
@@ -127,13 +163,28 @@ func (docker *Docker) ExposeUDP(ports ...int) *Docker {
...
+	portMappings = append(portMappings, PortMapping{
 		ContainerPort: containerPort,
 		HostPort:      hostPort,
 		ServicePort:   servicePort,
 		Protocol:      protocol})
+
```
timoreimann: nit-pick: Remove the line break.

timoreimann:
Review complete. Overall, I'm pretty happy with the change. Just a few details we need to discuss about.

kesselborn:
@timoreimann thanks a lot for your review!
I will correct the issues tomorrow morning

utils.go
kesselborn: ah ... left-over from converting pointers -> non-pointers

application.go
kesselborn: ack

application.go
kesselborn: yup

application.go
kesselborn: huh ... don't really know why I always made an extra variable ... I agree

application.go
kesselborn: you guessed right: I did this for consistency with the rest of this code. I would rather either do a linting PR or lint everything in this PR to minimize breaking changes PRs -- what do you think?

application_test.go
kesselborn: It's an example that will show up as an example in godoc. The reason I added an example here is the kind of weird interface marathon provides for contraints: It's an array of values which can differ in length. The hope is that an example clarifies this.

application.go
kesselborn: It's not used here anywhere but can be quite handy if you need to debug something using go-marathon as you can just pass it to Printf. I see `String()` as a textual representation of an object in a lose format (which is json here), so I wouldn't have a fixed expectation to test against.

application.go
kesselborn: no, but I tested to leave `Cmd` blank without `Args` being set and marathon did not complain. Depending on the executor, this could be valid behavior I guess. For the one I used, marathon executed `/bin/sh ""` which did not make a lot of sense -- however: the executor could be a different one I guess?

docker.go
kesselborn: mmm ... I think there are valid use cases for both scenarios:
  - I want to have everything empty and overwrite with empty values vs
  - I want to get an empty structure for updating and will only set values that I want to update
... the latter use case can be achieved by fetching the app from marathon before hand, the first one would mean that I manually have to set zero-values for all properties. If there will be a new property, lib users would have to adjust their code.

health.go
kesselborn: oh, you are right: via the ui you can't leave these fields empty, but via api it fills in default values

application.go
timoreimann: If by blank you mean empty string, then `/bin/sh [-c] ""` makes sense because the

> [cmd] value is wrapped by Mesos via `/bin/sh -c ${app.cmd}`.

(From the [POST docs](https://mesosphere.github.io/marathon/docs/rest-api.html#post-v2-apps).)

Not very helpful from a semantical point of view, but syntactically correct I guess. :-)

We run Docker containers in our Marathon/Mesos cluster exclusively, and more often than not leave both `cmd` and `args` empty. So I'm not sure what to make out of the Marathon documentation saying that _either `cmd` or `args` must be supplied_. Maybe it's set implicitly when using a Docker definition.

In that sense, I'm fine with your approach here.

application.go
timoreimann: I wouldn't lint everything now in order to keep the PR focused and drive towards shipping, but making a separate linting PR (or multiple ones to minimize breakage) sounds good to me.

So in favor of consistent names for now. :+1:

application.go
timoreimann: I can see how JSON as the format for an application configuration makes sense here. Testing would still be possible though as we could populate an `Application` instance with a set of  known values and verify that _just these_ were converted to the JSON structure we expect. It's certainly more effort than comparing to a complete known struct, however, and given that its main purpose is debugging/displaying content, I'm happy to leave out tests.

application_test.go
timoreimann: Haven't really worked with test examples yet, so that looks nice. :-)

docker.go
timoreimann: Not certain that I understand you completely here, partly because the term _empty_ is a bit ambiguous in our context. (Does it mean omitting as in the resulting JSON data, or zero-valued?)

Your first use case seems to be that of setting all values to zero, which doesn't sound like a common case as there's no inherent semantics in "everything zero-valued". The more common case to me is "I don't care about any value except for the following".

Having to fetch an application configuration prior to making changes to individual fields is at least theoretically prone to race conditions. Additionally, I'd rather not want go-marathon to reflect Marathon's default settings as these may be subject to change, causing things to break from one version to the next in the worst case. Finally, Marathon's `PUT` already has change-the-given-parameters-only semantics, so I'd prefer to use the existing mechanism.

That being said, I happen to realize now that things would be easier if the Marathon REST API had separate replace and update functionalities (i.e., real `PUT` and `PATCH`).

application.go
timoreimann: Still outstanding.

docker.go
timoreimann: Still outstanding.

application.go
kesselborn: yeah, that's what I meant. However: I _think_ the `/bin/sh -c ${app.cmd}` is just the default executor which can be changed using the `executor` property and then an empty value could be semantically correct.

application.go
kesselborn: done

application.go
kesselborn: done

application.go
kesselborn: done

kesselborn:
Today I did not have the chance to fix the remaining stuff ... just some notes to self as I lose track otherwise ;)

- [ ] https://github.com/gambol99/go-marathon/pull/127#discussion-diff-56251568
- [ ] https://github.com/gambol99/go-marathon/pull/127#discussion-diff-56251582
- [ ] https://github.com/gambol99/go-marathon/pull/127#discussion-diff-56251655
- [ ] https://github.com/gambol99/go-marathon/pull/127#discussion-diff-56251660

docker.go
kesselborn: done ... 
Additionally, I renamed the method from `Parameter` -> `AddParameter` to be more in line with other, similar methods. I can revert this though.

application.go
kesselborn: done

docker.go
kesselborn: done

docker.go
kesselborn: yeah ... I see your point. So: the only thing would be to initialize docker?

```go
func NewDockerContainer() *Container {
	container := &Container{}
	container.Type = "DOCKER"
	container.Docker = &Docker{}

	return container
}
```
That's what I have done now as initializing with `network: "BRIDGE"` would be inconsistent then.
Only gotcha here is that apart from this PR being a breaking change it would be a different behavior as well. But better now than later.

kesselborn:
Ok ... I am done with the next round ;)

docker.go
timoreimann: Agreed: A breaking change isn't ideal but for the better in this case as no default network should really be set.

One thing we could do to make people aware of the behavior shift is to mention the change in the commit message, maybe along with a **NOTE:** prefix.

timoreimann:
Looking pretty good at this point. I only have a few more suggestions to make:

1. Please fix the compile errors in `examples/` related to your API changes. (I was surprised to see that Travis didn't catch it; it's probably not configured to run tests across all packages.)
1. Please set the BRIDGED network in `examples/applications/main.go` so users have a working example for that interface.
1. Please reflect any API changes you made in the README file (e.g., `Arg()` -> `AddArgs()`).
1. Please set the BRIDGED network in the README file example too.
1. Update your initial commit message where needed. I'd remove the `Note: this is a request for comments` paragraph at least.
1. Feel free to add yourself to the AUTHORS file.
1. Squash all of your commits into one. Not a must, but since this PR has grown into a number of comment commits, would be nice to have some final git history clean-up.

kesselborn:
all done + squashed.

Note to self: unsquashed commit `6bec584`

timoreimann:
LGTM, I think we finally made it. :-)

Thanks again, Daniel. That was some major effort on your end!

kesselborn:
well: thank you for the repeated review effort -- glad I can use the upstream url again in my project :)

bobrik:
What was the reason for making `*map[string]string`? Is there a case where empty map has a different meaning than no key at all? Asking because:

```
invalid operation: a.Labels["foo"] (type *map[string]string does not support indexing)
```

timoreimann:
@bobrik The use case for having a pointer to a map (of labels) is that you may want to clear all labels for an application which previously had its labels set to something non-empty. You should use the convenience methods `AddLabel()` and `EmptyLabels()` to add and clear all labels, respectively.

Happy to hear if you can think of a way to simplify the design  while maintaining the given use cases.

bobrik:
@timoreimann I'm more worried about getting labels. I don't understand what prevents you from clearing all labels by making a new map without a pointer.

timoreimann:
@bobrik You'd need to dereference the map prior to indexing: `(*a.Labels)["foo"]`.

That being said, I think you're right that we should be able to use non-pointer maps. We somehow missed to see that the zero value for maps (and all other reference types, especially slices) is nil and becomes empty as soon as they are `make`d/initialized.

Do you agree that we should replace all pointer maps and slices by pure (non-pointer) maps and slices, respectively? I'm fine with putting together a PR myself or let you take it. Just let me know.

bobrik:
Yes, I think we should revert to using maps without pointers :+1:

PR #129, opened 2016-02-12
Improve AllTaskRunning, portIndex and added bool force to UpdateApplication

jglobant:
Hi,
we had to add the force flag to the update-method, so we can force rollbacks to previous version. It is the same like in the scale-method.
The problem with omitempty for json conversion was, that we wanted the portIndex in 0, so it uses the first port. But 0 was never sent to marathon due to the omitempty. There is another issue https://github.com/gambol99/go-marathon/pull/124 which deals with a similar problem.
The third change we had to made was a condition for healthcheck in AllTaskRunning. So if the app has health-checks, AllTasksRunning only returns true if all tasks are up && healthy.


timoreimann:
Frst of all: Ideally, the changes you propose should be broken down into three individual PRs instead of a big single one as they pertain to separate concerns. This would simplify discussions, evaluations, and other aspects.

You didn't extend or update any tests, so the test run fails as can be seen by the Travis CI report.

Regarding the individual features:

##### Force-update

Coincidentally, my team needed this feature as well, so we integrated it into our vendored copy of go-marathon. The plan was to file an upstream PR in the next few days -- you beat me to it. :)

As we have updated the tests already and did some refactorings to remove code duplication, how about you removing that part from your PR and let me submit [my branch](https://github.com/timoreimann/go-marathon/tree/support-forced-updates) instead?

##### portIndex

The Marathon REST API assumes `portIndex` to be 0, so there's no need to specify it explicitly. You've probably run into [this Marathon bug](https://github.com/mesosphere/marathon/issues/3140) that was introduced in 0.14. The 0.15.1 patch release already fixes this, and presumably the Marathon folks will fix it for 0.14 too. I don't know what others think about this, but I'd rather not change the go-marathon code base and rely on Marathon users to make a necessary system upgrade instead. Just my 2 cents.

Also note PR #127 that takes a pointer-based approach to dealing with zero value semantics. We should probably strive for a consistent way in any case.

##### health check

Can't comment on this one (yet): I'm not familiar with that part of Marathon.

jglobant:
@timoreimann thanks for your comments!
I have no idea how to assign different commits to different PRs

First of all sorry for not considering the tests, I will fix that and add it to the PR

**Force-update** - go ahead.
**portIndex** - I agree to wait on Marathon release
**health check** - well, this is the only one which is left

application.go
```diff
@@ -122,10 +122,11 @@ func (r *Application) AllTaskRunning() bool {
 	if r.Tasks == nil {
 		return false
 	}
-	if r.TasksRunning == r.Instances {
-		return true
+	if r.HasHealthChecks() {
+		return r.TasksRunning == r.Instances && r.TasksHealthy == r.Instances
```
timoreimann: The way I understand [Marathon health checks](https://mesosphere.github.io/marathon/docs/health-checks.html) is that every application has a health check, with the default one being Mesos' knowledge of the task state. In that sense, shouldn't we always have a properly set `r.TasksHealthy` value? If that's the case, then we should be able to just replace `r.TasksRunning` in the original if-expression by `r.TasksHealthy`, thus simplifying the entire if-clause to

```
if r.TasksHealthy == r.Instances {
  return true
}
```

I should be easy to get persuaded on this issue though as I have zero knowledge running Marathon application that do not come with a health check.

application.go
```diff
@@ -122,10 +122,11 @@ func (r *Application) AllTaskRunning() bool {
...
+	} else {
+		return r.TasksRunning == r.Instances
```
timoreimann: We can drop the `else` here and directly return the boolean expression as we're already returning in the `if` case. Or, if you agree to my reasoning above, directly go with

```
return r.TasksHealthy == r.Instances
```

timoreimann:
@jglobant: A PR is just a number of commits living on the same branch. Thus, if you bundle commits associated with a particular feature on the same branch, you can submit one PR per branch.

I have submitted PR #130 containing my implementation of the force-update feature. Feel free to have a look and comment.

By the way, the Mesosphere folks [said on Friday](https://github.com/mesosphere/marathon/issues/3140#issuecomment-183364499) that the bugfix release for 0.14 is on its way.

I also reviewed your change to `AllTaskRunning` and left some comments.

My suggestion would now be that you trim down your PR to just the `AllTaskRuning` part by undoing the changes for the force-updating and port index features. There are two ways to achieve this: either 

- create a new commit and push that commit to your branch; or
- modify the existing commit and force-push it to your branch.

I prefer the latter approach for commit history hygiene reasons. Feel free to do whatever you feel most comfortable with though.

Thanks!

application.go
jglobant: Hi @timoreimann 
not every app has healthchecks. Starting a simple app shows TasksHealthy in 0. That is why I added the condition with HasHealthChecks.

jglobant:
@timoreimann I will create a new PR with th HC isolated

PR #130, opened 2016-02-13
Add support for force-updating an application.

timoreimann:
This PR adds support for force-updating an application. REST-wise, it corresponds to passing the `force=true` query parameter to the `PUT /v2/apps` request.


gambol99:
:+1: LGTM

timoreimann:
Thanks for reviewing!

PR #131, opened 2016-03-03
Add support for fetch in Application

jhaals:
`fetch` is the replacement for `uris` which has been deprecated since 0.15.0

### DIFF
--- a/application.go
+++ b/application.go
@@ -67,6 +67,7 @@ type Application struct {
 	Labels                map[string]string   `json:"labels,omitempty"`
 	AcceptedResourceRoles []string            `json:"acceptedResourceRoles,omitempty"`
 	LastTaskFailure       *LastTaskFailure    `json:"lastTaskFailure,omitempty"`
+	Fetch                 []Fetch             `json:"fetch,omitempty"`
 }
 
 // ApplicationVersions is a collection of application versions for a specific app in marathon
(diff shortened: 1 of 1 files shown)


timoreimann:
Thanks for the PR!

If I am not mistaken though, we are missing out on a particular use case: We won't be able to specify that either `extract` or `cache` should be disabled. The reason is that Marathon, [according to the docs](https://mesosphere.github.io/marathon/docs/rest-api.html#post-v2-apps), sets the default for these parameters to `true`, and setting any of them to `false` in go-marathon would cause the parameter to get dropped in the resulting serialized JSON due to the specified `omitempty` tag. Effectively, we'd be back to the default value of `true` again.

This is a [known caveat when talking to REST APIs in Go](https://willnorris.com/2014/05/go-rest-apis-and-pointers), and a standard approach to handle this is to use pointers along with a set of convenience functions. In this case, however, I think we'd be just fine leaving out `omitempty` and rely on the user to choose the correct values.

Just if you like, you can go the extra mile and still provide a few helper functions (along with tests) to abstract away from the somewhat [inconvenient Mesos semantics](http://mesos.apache.org/documentation/latest/fetcher/). My suggestion:

- `Fetch(uri string, cache bool)`: Disables both `executable` and `extract`.
- `FetchExecutable(uri string, cache bool)`: Enables `executable` (and doesn't even bother setting `extract` because in this case, the "field is ignored and has no effect" anyway).
- `FetchExtractable(uri string, cache bool)`: Disables `executable` and enables `extract`.

The wrappers are just a nice bonus from my perspective. I'd be happy to accept things if we leave out `omitempty`. Someone else can always pick up from there and improve. :-)

jhaals:
Thank's for a detailed reply @timoreimann!
I've removed the `omitempty` tags.

timoreimann:
Thanks, LGTM!

@gambol99, @eicca: Any final thoughts? Or may I merge?

gambol99:
:+1: LGTM

PR #132, opened 2016-03-14
Add application configuration method

kesselborn:
This method returns a specific version of an application which can be helpful if you have running instances of application configurations which are not the current application configuration.

### DIFF
--- a/application.go
+++ b/application.go
@@ -359,7 +359,8 @@ func (r *marathonClient) Application(name string) (*Application, error) {
 func (r *marathonClient) ApplicationByVersion(name, version string) (*Application, error) {
 	var app *Application
 
-	if err := r.apiGet(fmt.Sprintf("%s/%s/versions/%s", marathonAPIApps, trimRootPath(name), version), nil, &app); err != nil {
+	uri := fmt.Sprintf("%s/versions/%s", buildURI(name), version)
+	if err := r.apiGet(uri, nil, &app); err != nil {
 		return nil, err
 	}
 
--- a/application.go
+++ b/application.go
@@ -353,10 +353,10 @@ func (r *marathonClient) Application(name string) (*Application, error) {
 	return wrapper.Application, nil
 }
 
-// ApplicationConfiguration retrieves the application configuration from marathon
+// ApplicationByVersion retrieves the application configuration from marathon
 // 		name: 		the id used to identify the application
 // 		version:  the version of the configuration you would like to receive
-func (r *marathonClient) ApplicationConfiguration(name, version string) (*Application, error) {
+func (r *marathonClient) ApplicationByVersion(name, version string) (*Application, error) {
 	var app *Application
 
 	if err := r.apiGet(fmt.Sprintf("%s/%s/versions/%s", marathonAPIApps, trimRootPath(name), version), nil, &app); err != nil {
--- a/application.go
+++ b/application.go
@@ -353,6 +353,19 @@ func (r *marathonClient) Application(name string) (*Application, error) {
 	return wrapper.Application, nil
 }
 
+// ApplicationConfiguration retrieves the application configuration from marathon
+// 		name: 		the id used to identify the application
+// 		version:  the version of the configuration you would like to receive
+func (r *marathonClient) ApplicationConfiguration(name, version string) (*Application, error) {
+	var app *Application
+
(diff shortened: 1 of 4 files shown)


gambol99:
LGTM ... any chance we could add a quick test

kesselborn:
Can do, yes. I thought about it yesterday, but as far as I can see I would add a new fixture and check if everything works? As it basically is the same Method as `Application` apart from the `app` property being root or second level element, I thought the only way this method will fail is if marathon changed its behavior, which again wouldn't be caught by this test.
Will do so anyways.

application.go
```diff
@@ -353,6 +353,19 @@ func (r *marathonClient) Application(name string) (*Application, error) {
 	return wrapper.Application, nil
 }
 
+// ApplicationConfiguration retrieves the application configuration from marathon
+// 		name: 		the id used to identify the application
+// 		version:  the version of the configuration you would like to receive
+func (r *marathonClient) ApplicationConfiguration(name, version string) (*Application, error) {
```
timoreimann: I'd rename this to `ApplicationByVersion` in order to be both more consistent with `Application` (that already retrieves a configuration and doesn't have the `Configuration` suffix) and distinguish from `ApplicationVersions` (which doesn't retrieve any configurations but just a list of versions). I think the Marathon naming is a bit unfortunate and ambiguous, so we should try to do better. :-)

application.go
```diff
@@ -353,6 +353,19 @@ func (r *marathonClient) Application(name string) (*Application, error) {
...
+	var app *Application
+
+	if err := r.apiGet(fmt.Sprintf("%s/%s/versions/%s", marathonAPIApps, trimRootPath(name), version), nil, &app); err != nil {
```
timoreimann: The `apiGet()` parameter passing can be simplified by employing `buildURI()` (similar to how the [ApplicationVersions implementation](https://github.com/gambol99/go-marathon/blob/de3f7b297e825ac61224ad3a3c0b88ba814a23cb/application.go#L321) does it) like this:

```
uri := fmt.Sprintf("%s/versions/%s", buildURI(name), version)
if err := r.apiGet(uri, nil, &app); err != nil {
```

application_test.go
```diff
@@ -294,3 +294,17 @@ func TestApplication(t *testing.T) {
...
+
+	application, err := endpoint.Client.ApplicationConfiguration(fakeAppName, "2014-09-12T23:28:21.737Z")
+	assert.NoError(t, err)
+	assert.NotNil(t, application)
+	assert.Equal(t, application.ID, fakeAppName)
+	assert.NotNil(t, application.HealthChecks)
+	assert.NotNil(t, application.Tasks)
+	assert.Equal(t, len(application.HealthChecks), 1)
+	assert.Equal(t, len(application.Tasks), 2)
+}
```
timoreimann: We should also test the unhappy path quickly by either one of two approaches:

1. Call the new method with with invalid name and verify that an error is returned (basically copying-and-pasting the very logic that [TestApplication() already has](https://github.com/gambol99/go-marathon/blob/master/application_test.go#L281-L282)).
1. Extract the test assertion logic common to `TestApplication` and your new test function into a non-exported helper function and call it from each test.

I prefer no. 2: While code redundancy isn't necessarily a bad thing in unit-testing, here I think it's actually desirable to get rid of it because we want all configuration-retrieving methods to return the same content.

timoreimann:
> As it basically is the same Method as Application apart from the app property being root or second level element, I thought the only way this method will fail is if marathon changed its behavior, which again wouldn't be caught by this test.

I believe we want a dedicated test to ensure that your implementation (which is separate from `Application()`) is correct now and stays correct when we do any refactorings. In that sense, it's not just Marathon which could break the new functionality but also me doing something stupid on my next PR. ;-)

(However, we could simplify the test if we implemented `ApplicationConfiguration()` in terms of a common implementation shared between that method and `Application()`.)

application.go
kesselborn: yeah, sounds better

application.go
kesselborn: ah, good one. I added this method on an older branch of `go-marathon`, that's why I missed this, thanks

application_test.go
kesselborn: yup, makes sense and I agree to extract the logic

kesselborn:
yeah, I admit: the reasoning was BS :D.
Addressed most of your points. Implementations don't share implementation as they are so short already that I think trying to extract something of them makes it more complex than simpler.

timoreimann:
Sounds reasonable.

LGTM from here. Merging now, thanks for contributing!

PR #134, opened 2016-03-19
Use default format verb when logging example marathon.Task.

timoreimann:
`marathon.Task` does not implement `String()`, so we replace %s by %v in `fmt.Printf()`. This fixes a `go vet` complain.


timoreimann:
Should be easy to review. @gambol99 maybe? Thanks.

gambol99:
apologizes for the delay ... easter holidays :-) ... LGTM

timoreimann:
No worries. Thanks!

PR #142, opened 2016-04-18
Support passing HTTP Auth credentials to event stream

anxueyan:
Support passing HTTP Auth credentials to event stream。

### DIFF
--- a/subscription.go
+++ b/subscription.go
@@ -166,7 +166,15 @@ func (r *marathonClient) registerSSESubscription() error {
 	url := fmt.Sprintf("%s/%s", marathon, marathonAPIEventStream)
 
 	// Try to connect to stream
-	stream, err := eventsource.Subscribe(url, "")
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return err
+	}
+	if r.config.HTTPBasicAuthUser != "" {
+		req.SetBasicAuth(r.config.HTTPBasicAuthUser, r.config.HTTPBasicPassword)
+	}
+
+	stream, err := eventsource.SubscribeWithRequest("", req)
 	if err != nil {
 		return err
 	}


subscription.go
```diff
@@ -166,7 +166,15 @@ func (r *marathonClient) registerSSESubscription() error {
 	url := fmt.Sprintf("%s/%s", marathon, marathonAPIEventStream)
 
 	// Try to connect to stream
-	stream, err := eventsource.Subscribe(url, "")
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return err
+	}
+	if r.config.HTTPBasicAuthUser != "" {
```
timoreimann: I'm not completely sure if HTTP Basic Auth supports empty usernames with non-empty passwords. When in doubt, we should probably extend the if condition by `|| r.config.HTTPBasicPassword != ""`.

timoreimann:
Would you mind also writing some tests for the new functionality (and the various branches it contains)?

Let me know if you need help.

timoreimann:
It also seems like the _Debug_ part in your commit message isn't perfectly fitting. How about changing it to _Support passing HTTP Auth credentials to event stream_ or something like that?

anxueyan:
@timoreimann Thanks for your advice。I will write some tests for my commit。

dmajere:
@anxueyan 
Thanks for your commit, I'm very interested in this PR, do you need any help with unit tests?

anxueyan:
@timoreimann  Sorry for long time to reply. Yes ,you are right, we should add if condition
 "|| r.config.HTTPBasicPassword != ". Even if HTTP credentials of marathon not support non-empty passwords or usernames It should allow pass empty passwords or usernames to marathon , the result may be authentication failed.  but I have two questions：

1.The file client.go,  It's also use the if condition "if r.config.HTTPBasicAuthUser != "" " in the function "func (r *marathonClient) apiCall(method, uri string, body, result interface{}) error " (line 247).  why this place use this condition？ 
2.  I run marathon in container with environment "MARATHON_HTTP_CREDENTIALS: "username:" "  or  "MARATHON_HTTP_CREDENTIALS: ":password" " , and do request like this 
"# curl -I -u kdkdjdksls:lkkd http://10.17.102.16:8089/v2/apps  (10.17.102.16:8089 is host:port of my marathon)"
the  response is "HTTP/1.1 200 OK" .  It's right? do you have met? It didn't seem to do certification .

anxueyan:
@dmajere yes, It's need some help with unit tests. I met the problem mentioned above， and I don't know how to write a good unit tests. Here is an example, is what I write：
```
  func TestEventStreamConnectionHTTPNullAuthUser(t *testing.T) {
	config := NewDefaultConfig()
	config.EventsTransport = EventsTransportSSE
	config.URL = "http://non-existing-marathon-host.local:5555"
	config.HTTPBasicAuthUser = ""
	config.HTTPBasicPassword = "password"

	endpoint := newFakeMarathonEndpoint(t, &config)
	defer endpoint.Close()

	events := make(EventsChannel)
	err := endpoint.Client.AddEventsListener(events, EVENTS_APPLICATIONS)

	assert.NoError(t, err)
	endpoint.Client.RemoveEventsListener(events)
}
```
It's ok? if not ok ,how should I write?

dmajere:
@anxueyan 
In your case, as i see it, we need to verify that req object method [BasicAuth](https://golang.org/pkg/net/http/#Request.BasicAuth) returns user, passwd when we set them.
To do this we'll need to substitute eventsource.SubscribeWithRequest method, so we'll be able to record parameters passed to it and validate them.
To do this we'll have to change registerSSESubscription() method to registerSSESubscription(dialer) where dialer is a function or an interface, this way, we'll be able to pass eventsource.SubscribeWithRequest in our library and mock object in our tests.

timoreimann:
@anxueyan Regarding your questions:

1. The [section in question](https://github.com/gambol99/go-marathon/blob/52466f39391c90839b2a75715d5c5af68b7e7502/client.go#L247) passes the credentials defined during the creation of the go-marathon client configuration to Marathon on every HTTP request. If you're wondering why we have to do this again in this PR: It's because event streams are implemented through an external library which doesn't run through go-marathon's low-level `apiCall()` method. It's good that you spotted this though as it seems to me that we should extract the existing logic into a utility function in order to be applied from both call sites.
2. I'm not entirely sure, but maybe this indicates that password-less credentials are in fact not possible with Marathon? Then again, I'd be surprised if Marathon doesn't return an error in this case. Maybe you can find something in the logs?

anxueyan:
@timoreimann The first question， I am confused why there is not add if condition
 ```|| r.config.HTTPBasicPassword != ""```

timoreimann:
1. I suppose the condition is missing because nobody thought about the use case of an empty username prior to this PR. :-). That's why I suggested to extract the logic into a central function so that we avoid making the same mistakes at multiple call sites across the code base. Although in this case...
2. It turned out now that you must always pass both username and password anyway. Consequently, we don't need to extend the logical operation. However, I still think it'd be useful to centralize the credentials logic on the request instance for future changes on that part to come.

anxueyan:
@timoreimann It's better to centralize the credentials logic on the request instance . 😊

timoreimann:
@anxueyan The request instance is from the stdlib (`http.Request`) which cannot be amended. A simple wrapper function like `setCredentials(req *http.Request, c *config)` should hopefully do the trick. :-)

timoreimann:
@anxueyan did you have a chance to try out @dmajere's testing approach? Let us know if there's need for clarification, and feel free to push any intermediate results you may have at any point in time. Thanks!

gambol99:
@timoreimann  @anxueyan  ... Since its very much related to issue #148 (auth+tls) .. and the PR https://github.com/donovanhide/eventsource/pull/22 has been merged, perhaps we can clean both up with the below ... or similar ..
```go
url := fmt.Sprintf("%s/%s", marathon, marathonAPIEventStream)
// step: create a new request for the stream
request, err := http.NewRequest("GET", url, nil)
if err != nil {
	return err
}
// step: add any basic authentication (perhaps wrap this in a method)
if r.config.HTTPBasicAuthUser != "" && r.config.HTTPBasicPassword != "" {
	request.SetBasicAuth(r.config.HTTPBasicAuthUser, r.config.HTTPBasicPassword)
}
// Try to connect to stream, reusing the http client settings
stream, err := eventsource.SubscribeWith("", r.httpClient, request)
if err != nil {
	return err
}
```

gambol99:
was fixed in PR https://github.com/gambol99/go-marathon/pull/156

PR #143, opened 2016-04-28
Fix golint errors in events.go and its consumers.

### LINKED ISSUE #80: Golint conformance

timoreimann:
**BREAKING CHANGE PSA:** This commit changes all public event ID names and a few struct field names. Consumers will need to adjust, unfortunately.

Closes #80.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ if err != nil {
 
 // Register for events
 events := make(marathon.EventsChannel, 5)
-err = client.AddEventsListener(events, marathon.EVENTS_APPLICATIONS)
+err = client.AddEventsListener(events, marathon.EventIDApplications)
 if err != nil {
 	log.Fatalf("Failed to register for events, %s", err)
 }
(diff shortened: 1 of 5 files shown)


timoreimann:
@eicca would you like to review?

timoreimann:
Pinging @gambol99 as potential reviewer too just in case :-)

gambol99:
absolutely! ... always hated those .. and i was the one who did it :-) ... LGTM

timoreimann:
Thanks ☺️

PR #147, opened 2016-05-12
marathon task.Ipaddresses field added

dmajere:
Good day,
since marathon version 1.1.1 additional field in Task structure has been introduced.
It look like this.
`            
"ipAddresses": [
                {
                    "ipAddress": "0.0.0.0",
                    "protocol": "IPv4"
                }
            ]
`

### DIFF
--- a/events.go
+++ b/events.go
@@ -129,15 +129,16 @@ type EventAPIRequest struct {
 
 // EventStatusUpdate describes a 'status_update_event' event.
 type EventStatusUpdate struct {
-	EventType  string `json:"eventType"`
-	Timestamp  string `json:"timestamp,omitempty"`
-	SlaveID    string `json:"slaveId,omitempty"`
-	TaskID     string `json:"taskId"`
-	TaskStatus string `json:"taskStatus"`
-	AppID      string `json:"appId"`
-	Host       string `json:"host"`
-	Ports      []int  `json:"ports,omitempty"`
-	Version    string `json:"version,omitempty"`
+	EventType   string       `json:"eventType"`
+	Timestamp   string       `json:"timestamp,omitempty"`
+	SlaveID     string       `json:"slaveId,omitempty"`
+	TaskID      string       `json:"taskId"`
+	TaskStatus  string       `json:"taskStatus"`
+	AppID       string       `json:"appId"`
+	Host        string       `json:"host"`
+	Ports       []int        `json:"ports,omitempty"`
+	IPAddresses []*IPAddress `json:"ipAddresses"`
+	Version     string       `json:"version,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


dmajere:
I found out that ipAddresses field is also appeared in Event structure, but I don't know where is a better place for IpAddress struct, should I create ipam.go file for it? Any suggestions?

timoreimann:
@dmajere I assume that tasks can be considered the real owner of this new struct, whereas events simply "consume" it. In that sense, I'd keep it in `task.go` and just reference from whereever needed. In fact, already do this in `event.go` for other structs.

Could you please point me to the Marathon documentation of this new struct? Thanks!

dmajere:
It seems that there is no information about it in documentation yet.
I found this field first in marathon dashboard.
Here are code snippets from marathon with classes for status update event and launched task.
 
src/main/scala/mesosphere/marathon/event/Events.scala
`
221 case class MesosStatusUpdateEvent(
222   slaveId: String,
223   taskId: Task.Id,
224   taskStatus: String,
225   message: String,
226   appId: PathId,
227   host: String,
228   ipAddresses: Option[Seq[org.apache.mesos.Protos.NetworkInfo.IPAddress]],
229   ports: Seq[Int],
230   version: String,
231   eventType: String = "status_update_event",
232   timestamp: String = Timestamp.now().toString) extends MarathonEvent`

src/main/scala/mesosphere/marathon/core/task/Task.scala
`
408   case class Launched(
409       appVersion: Timestamp,
410       status: Status,
411       hostPorts: Seq[Int]) {
412
413     def hasStartedRunning: Boolean = status.startedAt.isDefined
414
415     def ipAddresses: Option[Seq[MesosProtos.NetworkInfo.IPAddress]] =
416       status.mesosStatus.flatMap(MesosStatus.ipAddresses)
417   }
`

timoreimann:
@dmajere Thanks for the pointers.

I overlooked a tiny thing when I reviewed your PR: Could you please upper-case the initialisms in `IpAddress[es]` to `IPAddress[es]` in order to satisfy the [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)?

Unfortunately, I don't have a Marathon 1.1.1 copy around right now. Did you possibly test and verify that the struct is populated properly?

dmajere:
>  Could you please upper-case the initialisms in IpAddress[es] to IPAddress[es] in order to satisfy the Go Code Review Comments?

Done.

We've been using go-marathon with this patch for almost a week now and got no problems so far.

timoreimann:
Great to hear. Sounds like it shouldn't cause any problems then.

LGTM. Thanks for contributing!

Issue #110, opened 2015-12-06
Add tests for WaitOnApplication


timoreimann:
The method in question does not seem to be covered by tests. We should do that.

PR #149, opened 2016-05-19
Issue110: Add tests for WaitOnApplication

dmajere:
#110 
Add tests for WaitOnApplication

### DIFF
--- a/application_test.go
+++ b/application_test.go
@@ -18,6 +18,7 @@ package marathon
 
 import (
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -426,4 +427,23 @@ func TestApplicationConfiguration(t *testing.T) {
 	apiErr, ok = err.(*APIError)
 	assert.True(t, ok)
 	assert.Equal(t, ErrCodeNotFound, apiErr.ErrCode)
+}
+
+func TestWaitOnApplication(t *testing.T) {
+	endpoint := newFakeMarathonEndpoint(t, nil)
+	defer endpoint.Close()
+
+	var err error
+	err = endpoint.Client.WaitOnApplication(fakeAppName, 1*time.Second)
+	assert.NoError(t, err)
+
+	err = endpoint.Client.WaitOnApplication("no_such_app", 1*time.Millisecond)
+	assert.IsType(t, err, ErrTimeoutError)
+}
+func TestAppExistAndRunning(t *testing.T) {
+	endpoint := newFakeMarathonEndpoint(t, nil)
+	defer endpoint.Close()
+	client := endpoint.Client.(*marathonClient)
+	assert.True(t, client.appExistAndRunning(fakeAppName))
+	assert.False(t, client.appExistAndRunning("no_such_app"))
 }
--- a/application.go
+++ b/application.go
@@ -545,31 +545,35 @@ func (r *marathonClient) CreateApplication(application *Application) (*Applicati
 //		name:		the id of the application
 //		timeout:	a duration of time to wait for an application to deploy
 func (r *marathonClient) WaitOnApplication(name string, timeout time.Duration) error {
-	// step: this is very naive approach - the problem with using deployment id's is
-	// one) from > 0.8.0 you can be handed a deployment Id on creation, but it may or may not exist in /v2/deployments
-	// two) there is NO WAY of checking if a deployment Id was successful (i.e. no history). So i poll /deployments
-	// as it's not there, was it successful? has it not been scheduled yet? should i wait for a second to see if the
-	// deployment starts? or have i missed it? ...
-	err := deadline(timeout, func(stop_channel chan bool) error {
-		var flick atomicSwitch
-		go func() {
-			<-stop_channel
-			close(stop_channel)
-			flick.SwitchOn()
-		}()
-		for !flick.IsSwitched() {
-			app, err := r.Application(name)
-			if apiErr, ok := err.(*APIError); ok && apiErr.ErrCode == ErrCodeNotFound {
-				continue
-			}
-			if err == nil && app.AllTaskRunning() {
+	if r.appExistAndRunning(name) {
+		return nil
+	}
+
+	ticker := time.NewTicker(time.Millisecond * 500)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-time.After(timeout):
+			return ErrTimeoutError
+		case <-ticker.C:
+			if r.appExistAndRunning(name) {
 				return nil
 			}
-			time.Sleep(time.Duration(500) * time.Millisecond)
 		}
-		return nil
-	})
-	return err
+	}
+	return nil
+}
+
+func (r *marathonClient) appExistAndRunning(name string) bool {
+	app, err := r.Application(name)
+	if apiErr, ok := err.(*APIError); ok && apiErr.ErrCode == ErrCodeNotFound {
+		return false
+	}
+	if err == nil && app.AllTaskRunning() {
+		return true
+	}
+	return false
 }
 
 // DeleteApplication deletes an application from marathon


application_test.go
```diff
@@ -427,3 +428,21 @@ func TestApplicationConfiguration(t *testing.T) {
 	assert.True(t, ok)
 	assert.Equal(t, ErrCodeNotFound, apiErr.ErrCode)
 }
+
+func TestWaitOnApplication(t *testing.T) {
+	endpoint := newFakeMarathonEndpoint(t, nil)
+	defer endpoint.Close()
+
+	err := endpoint.Client.WaitOnApplication(fakeAppName, 0*time.Millisecond)
```
timoreimann: I'm wondering if this approach carries a risk for flakiness: It basically relies on the timeout error returning before the result of `TestWaitOnApplication`'s does. On my local machine, the test would always succeed; I'm not sure though if a different machine or system behavior could make a change.

Any thoughts on this one?

application_test.go
```diff
@@ -427,3 +428,21 @@ func TestApplicationConfiguration(t *testing.T) {
...
+	if assert.Error(t, err) {
```
timoreimann: I think you should be able to skip this one and use `assert.EqualError` directly.

application_test.go
```diff
@@ -427,3 +428,21 @@ func TestApplicationConfiguration(t *testing.T) {
...
+		assert.EqualError(t, err, "the operation has timed out")
+	}
+
+	err = endpoint.Client.WaitOnApplication("no_such_app", 5*time.Millisecond)
+	if assert.Error(t, err) {
```
timoreimann: Same here: Try using `assert.EqualError` directly.

application_test.go
```diff
@@ -427,3 +428,21 @@ func TestApplicationConfiguration(t *testing.T) {
...
 	assert.Equal(t, ErrCodeNotFound, apiErr.ErrCode)
 }
+
+func TestWaitOnApplication(t *testing.T) {
+	endpoint := newFakeMarathonEndpoint(t, nil)
+	defer endpoint.Close()
+
+	err := endpoint.Client.WaitOnApplication(fakeAppName, 0*time.Millisecond)
+	if assert.Error(t, err) {
+		assert.EqualError(t, err, "the operation has timed out")
```
timoreimann: Compare with [ErrTimeoutError](https://github.com/gambol99/go-marathon/blob/a78e3aa5d5ca7d5b14eafc3239d61bb53af17434/client.go#L144) directly.

application_test.go
```diff
@@ -427,3 +428,21 @@ func TestApplicationConfiguration(t *testing.T) {
...
+	}
+
+	err = endpoint.Client.WaitOnApplication("no_such_app", 5*time.Millisecond)
+	if assert.Error(t, err) {
+		assert.EqualError(t, err, "the operation has timed out")
```
timoreimann: Compare with [ErrTimeoutError](https://github.com/gambol99/go-marathon/blob/a78e3aa5d5ca7d5b14eafc3239d61bb53af17434/client.go#L144) directly.

timoreimann:
Thanks for adding the missing test. I left some comments.

application_test.go
dmajere: @timoreimann 
Yep, my bad.
The better way is to extract anonymous function from WaitOnApplication method into separate one, so that we can isolate logic we want to test from side effects of deadline method.

dmajere:
I've extracted waitOnApplication and remade tests for it

application_test.go
```diff
@@ -427,3 +427,20 @@ func TestApplicationConfiguration(t *testing.T) {
...
+	client := endpoint.Client.(*marathonClient)
+
+	c := make(chan bool)
+	err := client.waitOnApplication(fakeAppName, c)
+	assert.NoError(t, err)
+
+	c = make(chan bool)
+	go func() {
+		c <- true
+	}()
```
timoreimann: That's pretty good already. I think we need to do one last thing though:

Unless I am mistaken, we still cannot be certain that we will make it into one iteration of `waitOnApplication`'s for-loop before the stop signal arrives. How about sleeping half a second inside this go function to give the the system-under-test enough time to hit a 404?

application_test.go
```diff
@@ -427,3 +427,20 @@ func TestApplicationConfiguration(t *testing.T) {
...
+	err = client.waitOnApplication("no_such_app", c)
+	assert.IsType(t, err, ErrTimeoutError)
```
timoreimann: I suppose `assert.EqualError` didn't work here for some reason?

dmajere: `assert.EqualError` works with strings, but I can change assertion to
 `assert.EqualError(t, err, ErrTimeoutError.String())` if you think it will be clearer

timoreimann: No, I think it's okay then. Better to compare by types if we can. 👍

application_test.go
dmajere: I played with this function once more, and I wonder, is there a reason why custom timer function (`deadline`) is used there?
I ended up with this 

```go
func (r *marathonClient) WaitOnApplication(name string, timeout time.Duration) error {

// we have to run this for the first time
    if r.appExistAndRunning(name) {
        return nil
    }

    deadline := time.NewTimer(timeout).C
    ticker := time.NewTicker(time.Millisecond * 500).C
    for {
        select {
        case <-deadline:
            return ErrTimeoutError
        case <-ticker:
            if r.appExistAndRunning(name) {
                return nil
            }
        }
    }
    return nil
    }
func (r *marathonClient) appExistAndRunning(name string) bool {

    app, err := r.Application(name)
    if apiErr, ok := err.(*APIError); ok && apiErr.ErrCode == ErrCodeNotFound {
        return false
    }
    if err == nil && app.AllTaskRunning() {
        return true
    }
    return false}
```

with the test
```go
func TestWaitOnApplication(t *testing.T) {
    endpoint := newFakeMarathonEndpoint(t, nil)
    defer endpoint.Close()
    client := endpoint.Client.(*marathonClient)

    err := client.WaitOnApplication(fakeAppName, 1*time.Second)
    assert.NoError(t, err)

    err = client.WaitOnApplication("no_such_app", 1*time.Second)
    assert.IsType(t, err, ErrTimeoutError)
}
func TestAppExistAndRunning(t *testing.T) {
    endpoint := newFakeMarathonEndpoint(t, nil)
    defer endpoint.Close()
    client := endpoint.Client.(*marathonClient)
    assert.True(t, client.appExistAndRunning(fakeAppName))
    assert.False(t, client.appExistAndRunning("no_such_app"))
}
```

What do you think about it?

timoreimann: I think your approach is certainly easier to read and understand. The test now seems non-racey as well.

The only real open question to me is whether we need the `atomicSwitch` for some reason. @gambol99, you created the code in the first place. Can you shed some light on whether @dmajere's implementation is equivalent to the previous code from your perspective?

Assuming it is, I have a few comments:

- You have to stop the Ticker in order to release the associated resources (see [docs](https://golang.org/pkg/time/#NewTicker)). So store the ticker in a variable, call `defer ticker.Stop()`, and read from the channel in the select statement.
- The documentation isn't clear on whether one needs to stop the timer as well. I'm thinking that you don't have to because the timer can clean up after itself once it has fired its single event. Not sure about this though.
- Assuming that we don't need to stop the timer: I think you can simplify a bit by dropping the `ticker` variable definition and replace `case <-ticker` with `case <-After(time.Millisecond * 500)`.

dmajere: @gambol99, I think `atomicSwitch` was used just for synchronisation of two goroutines, so usage of blocking channel would be more go-stylish.

@timoreimann, You seems to be right about timer cleaning up after firing up an event.
I've changed implementation to using `After()`. I'll push changes if switching from `atomicSwitch` is ok.

timoreimann: Feel free to push right away. (Just don't amend and force-push to ease reviewability.) We can continue to discuss on a latest commit (and should be easier too).

dmajere: Ok, done.

timoreimann:
LGTM pending @gambol99's judgement.

application_test.go
gambol99: yep @dmajere code looks functionally identical, but far easier to read ... 

> The only real open question to me is whether we need the atomicSwitch for some reason. > @gambol99, you created the code in the first place. Can you shed some light on whether > @dmajere's implementation is equivalent to the previous code from your perspective?

As for my poor atomSwitch implementation, my only defense is go-marathon was my first stab at writing go :-)

gambol99:
LGTM

dmajere:
Ok, I'll do commit squash before merge.

timoreimann:
Merged -- thanks for the test and the refactoring for the better! :-)

timoreimann:
@gambol99 I just realized that we left out the explanatory  comments at the beginning of the old `WaitOnApplication` implementation. Shall we make another commit to bring the comment back? Do you know if that statement still holds?

gambol99:
@timoreimann ... I think the comment was left from long before, when it was using deployment Id to check the status of a deployed application, rather than now, which is simply polling for the existence of the app by name.

timoreimann:
@gambol99 :+1:

Issue #111, opened 2015-12-09
kill task does not work if app id has  hierarchical path


lokeshverizon:
```
func (r *marathonClient) KillTask(taskID string, scale bool) (*Task, error) {
  ...
  appNamewithUnderScore := taskID[0:strings.LastIndex(taskID, ".")]
       
  // following line needs to be added. 
  appName := strings.Replace(appNamewithUnderScore, "_", "/", -1)
}
```

timoreimann:
If I understand you correctly, the issue is about Marathon app names containing slashes that would end up in task IDs again. Looking at the [official REST documentation](https://mesosphere.github.io/marathon/docs/rest-api.html#post-v2-apps) (see the _identifier_ section), however, such Marathon IDs should be rejected in the first place as the slash is not among the set of allowed characters.

Am I missing something?

lokeshverizon:
we can have app id with hierachical name. /group/project/application. 
Hence appid = /group/project/application
marathon is able to launch this application. this is converted to mesostask as:
group_project_application.<taskid>
with current API: it gives: group_project_application
which fails.

lokeshverizon:
also in the doc you mentioned:     "id": "/product/service/myApp",

which is similar case as mine.

lokeshverizon:
let me know if you need more info. i have local fix for this now. 
 appName := strings.Replace(appNamewithUnderScore, "_", "/", -1)
but would prefer to get it fix here.

timoreimann:
Taking over this one.

@lokeshverizon, are you still interested in fixing this upstream? I'd be happy to review any pull requests you submit

dmajere:
When you're trying to create marathon application, it will split your AppID into segments by "/", and then check if every segment is matching 
`^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])$|^(\\.|\\.\\.)$`

I've been thinking about implementing the same AppID validation in go-marathon library, so it can be run in app.Name() method and Application constructors. 
We have utils.validateID(), but it is not really validating ID.
I suggest changing
 `validateID(string) string`
 to 
`validateID(string) (bool, string)` and add actual validation in it.

dmajere:
This one is solved with #150

PR #150, opened 2016-05-19
Issue111: KillTask does not work if AppID has hierarchical path

dmajere:
This one is intend to solve #111 
It's worth to mention that marathon is checking if AppID is matching the following regexp
`
^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])|(\\.|\\.\\.)$
`
Basically it means that _ symbol is not allowed, and it is used as a delimiter between group/subgroup ids and AppID  in TaskID. So I've changed all app ids in test fixture from fake_app to fake-app because it was invalid.

### DIFF
--- a/task.go
+++ b/task.go
@@ -116,6 +116,9 @@ func (r *marathonClient) KillApplicationTasks(id string, opts *KillApplicationTa
 //	opts:		KillTaskOpts request payload
 func (r *marathonClient) KillTask(taskID string, opts *KillTaskOpts) (*Task, error) {
 	appName := taskID[0:strings.LastIndex(taskID, ".")]
+	appName = strings.Replace(appName, "_", "/", -1)
+	taskID = strings.Replace(taskID, "/", "_", -1)
+
 	u := fmt.Sprintf("%s/%s/tasks/%s", marathonAPIApps, appName, taskID)
 	u, err := addOptions(u, opts)
 	if err != nil {
--- a/task_test.go
+++ b/task_test.go
@@ -67,13 +67,21 @@ func TestKillApplicationTasks(t *testing.T) {
 }
 
 func TestKillTask(t *testing.T) {
+	cases := map[string]struct {
+		TaskID string
+		Result string
+	}{
+		"CommonApp":           {fakeTaskID, fakeTaskID},
+		"GroupApp":            {"fake-group_fake-app.fake-task", "fake-group_fake-app.fake-task"},
+		"GroupAppWithSlashes": {"fake-group/fake-app.fake-task", "fake-group_fake-app.fake-task"},
+	}
 	endpoint := newFakeMarathonEndpoint(t, nil)
 	defer endpoint.Close()
 
-	task, err := endpoint.Client.KillTask(fakeTaskID, nil)
-	assert.NoError(t, err)
-	if assert.NotNil(t, task) {
-		assert.Equal(t, fakeTaskID, task.ID)
+	for k, tc := range cases {
+		task, err := endpoint.Client.KillTask(tc.TaskID, nil)
+		assert.NoError(t, err, "TestCase: %s", k)
+		assert.Equal(t, tc.Result, task.ID, "TestCase: %s", k)
 	}
 }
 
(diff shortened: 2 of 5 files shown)


gambol99:
LGTM ...

task.go
```diff
@@ -116,6 +116,8 @@ func (r *marathonClient) KillApplicationTasks(id string, opts *KillApplicationTa
 //	opts:		KillTaskOpts request payload
 func (r *marathonClient) KillTask(taskID string, opts *KillTaskOpts) (*Task, error) {
 	appName := taskID[0:strings.LastIndex(taskID, ".")]
+	appName = strings.Replace(appName, "_", "/", -1)
```
timoreimann: Do we only need the replacement in this case? What about other calls to Marathon? I'm not saying that we need to fix everything along this PR, just that we should create a ticket if there's more sanitization to do.

Did you add/extend a test for the replacement?

dmajere: @timoreimann, I wanted to to this
Its a quote form related issue #111 

> When you're trying to create marathon application, it will split your AppID into segments by "/", and then check if every segment is matching 
^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])$|^(\\.|\\.\\.)$
> 
> I've been thinking about implementing the same AppID validation in go-marathon library, so it can be run in app.Name() method and Application constructors. 
We have utils.validateID(), but it is not really validating ID.
I suggest changing
validateID(string) string
to 
validateID(string) (bool, string) and add actual validation in it.

It seem's that other methods working fine. But I'll check it out once more.

dmajere: Yep, it all other methods are working properly

timoreimann: Sorry, I missed the comment in the related issue. Doesn't this mean that a call to `CreateApplication()` will fail as well if the given app ID is invalid?

Again, I'm not keen of getting this fixed right now as the issue and PR are about `KillTask`. Just want to make sure we document things we discover along the way.

timoreimann: Aren't we yet missing a test that passes an underscored app ID into `KillTask` and makes sure we're seeing the proper replacement?

dmajere: >  Doesn't this mean that a call to CreateApplication() will fail as well if the given app ID is invalid?

Yes, it does. Actually it does now to. But the difference is that for now its done after sending request, I'm proposing to do it on method run instead.

> Aren't we yet missing a test that passes an underscored app ID into KillTask and makes sure we're seeing the proper replacement?

Yes, it was missed. I've changed test case to table driven style, added a fixture and a test case for `TaskKill` for application in group

timoreimann: I'm not entirely sure whether we should duplicate the ID validation logic in the library, my reasoning here being that it will increase coupling and require us to watch out for changes in this regard. It wouldn't be the first unannounced breaking change that the Marathon project does.

That's just my opinion though. Feel free to open a new issue on the subject to discuss.

task_test.go
```diff
@@ -67,13 +67,23 @@ func TestKillApplicationTasks(t *testing.T) {
...
 
-	task, err := endpoint.Client.KillTask(fakeTaskID, nil)
-	assert.NoError(t, err)
-	if assert.NotNil(t, task) {
-		assert.Equal(t, fakeTaskID, task.ID)
+	for k, tc := range cases {
+		task, err := endpoint.Client.KillTask(tc.TaskID, nil)
+		assert.NoError(t, err, "TestCase: %s", k)
+		if assert.NotNil(t, task, "TestCase: %s", k) {
+			assert.Equal(t, tc.Result, task.ID, "TestCase: %s", k)
```
timoreimann: Do we need the preceding check for nilness? I'd think that `Equal` is smart enough to check for it as well.

timoreimann:
Left one more comment on the new test. Everything else LGTM now!

task_test.go
dmajere: It was this way before my changes, I did not change assertion logic, but I think you're right. 
Fixed.

timoreimann:
Ready for merge. Thanks!

PR #154, opened 2016-05-30
fixed ping path to make Ping() work

Shtsh:
go-marathon tries to make ping request to http://marathon//ping
This commit removes redundant slash


timoreimann:
Awww, extra path component slashes: Usually works, but always looks ugly. :-)

Thanks a lot! :+1:

PR #155, opened 2016-06-04
Always encode HTTP error response content into API error instances

### LINKED ISSUE #151: Odd Error object issue
My app was having Marathon connectivity issues and unexpectedly panicked when using the `err` object  returned by `client.CreateApplication`.

My code:

```
        _, err := m.Client.CreateApplication(application)
	if err != nil {
		log.Error(err)
		if err.Error() == ErrClusterConnectivity.Error() {
			m.resetClient()
		}
      }
```

The call to `log.Error(err)` prints '`<nil>`' and the call to `err.Error()` panics with a null pointer exception. Since this is all wrapped by `if err != nil`, it feels like the err object returned by `.CreateApplication` (https://github.com/gambol99/go-marathon/blob/master/application.go#L538) might be improperly initialized.

This the actual line that causes the panic:  https://github.com/gambol99/go-marathon/blob/master/error.go#L55

Any idea what's going on here?

### LINKED ISSUE #152: Failure to parse content from error response causes nil error to be returned to caller
When [parseContent()](https://github.com/gambol99/go-marathon/blob/52466f39391c90839b2a75715d5c5af68b7e7502/error.go#L85) fails, it returns a `(nil, err)` pair to the [call site](https://github.com/gambol99/go-marathon/blob/52466f39391c90839b2a75715d5c5af68b7e7502/client.go#L280). We log the parse error dutifully but also [return](https://github.com/gambol99/go-marathon/blob/52466f39391c90839b2a75715d5c5af68b7e7502/client.go#L285) the (nil) `APIError` copy straight to the user, giving the impression that nothing failed (which is not correct).

We should return an embedded parser error so that callers may at least be able to see the unparsable error.

timoreimann:
This commit comprises two major changes:

- Instead of returning nil on non-JSON-parsable content from HTTP error  responses, fall back to encoding the content as-is.
- For unknown errors, try to parse the content as well.

Effectively, we always return a (non-nil) APIError instance and thus can simplify NewAPIError's signature by dropping the second (parse) error return value. We also stop returning nil error values that do not equal to nil, a problem known to [Go](https://golang.org/doc/faq#nil_error). We also take the opportunity to simplify the error-identifying switch statement.

Finally, we convert the existing error parsing tests into table-driven ones to increase readability and be more in-line with the conventional Go testing style.

Fixes #151.
Fixes #152.

### DIFF
--- a/client.go
+++ b/client.go
@@ -277,12 +277,7 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 		return nil
 	}
 
-	apiErr, err := NewAPIError(response.StatusCode, respBody)
-	if err != nil {
-		r.debugLog.Printf("apiCall(): failed to parse error response '%s' with status code %d, error: %s", respBody, response.StatusCode, err)
-	}
-
-	return apiErr
+	return NewAPIError(response.StatusCode, respBody)
 }
 
 var oneLogLineRegex = regexp.MustCompile(`(?m)^\s*`)
--- a/error.go
+++ b/error.go
@@ -57,24 +57,27 @@ func (e *APIError) Error() string {
 
 // NewAPIError creates a new APIError instance from the given response code and content.
 func NewAPIError(code int, content []byte) error {
+	var errDef errorDefinition
 	switch {
 	case code == http.StatusBadRequest:
-		return parseContent(&badRequestDef{}, content)
+		errDef = &badRequestDef{}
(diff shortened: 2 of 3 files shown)


timoreimann:
@gambol99 would you like to review?

/cc @iderdik

gambol99:
LGTM

timoreimann:
Thanks for reviewing!

Issue #148, opened 2016-05-12
Basic Auth/TLS broken in marathonClient.AddEventsListener()


emilevauge:
When we create a marathon client with basic auth and/or TLS, the call to `marathonClient.AddEventsListener()` returns an error:

```
401: <html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 401 </title>
</head>
<body>
<h2>HTTP ERROR: 401</h2>
<p>Problem accessing /v2/events. Reason:
<pre>    Unauthorized</pre></p>
<hr /><a href="http://eclipse.org/jetty">Powered by Jetty:// 9.3.z-SNAPSHOT</a><hr/>
</body>
</html>
```

with TLS:

```
x509: certificate is valid for *.mydomain.net, mydomain.net, not ***.***.***.***.mydomain.net
```
Here is the code used to register listeners: https://github.com/containous/traefik/blob/master/provider/marathon.go#L67

It seems marathonClient.AddEventsListener() doesn't use TLS/basic auth client config.

arichardet:
Any resolution on this?

timoreimann:
@arichardet I suppose PR #142 will provide a fix for the problem once it has been merged.

gambol99:
@timoreimann ... i think #142 fixes the auth but not TLS. I've raised a PR https://github.com/donovanhide/eventsource/pull/22 to permit overriding the client

timoreimann:
Good point, those are indeed two different issues. Thanks!

gambol99:
was fixed in #156

emilevauge:
Thanks a lot @gambol99 :)

PR #156, opened 2016-06-05
basic auth and http client settings

gambol99:
- adding basic auth and http client settings to the event stream client

### DIFF
--- a/client.go
+++ b/client.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net/http"
@@ -152,7 +153,7 @@ type marathonClient struct {
 	subscribedToSSE bool
 	// the ip address of the client
 	ipAddress string
-	// the http server */
+	// the http server
 	eventsHTTP *http.Server
 	// the http client use for making requests
 	httpClient *http.Client
@@ -226,7 +227,6 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 	if err != nil {
 		return err
 	}
-
 	url := fmt.Sprintf("%s/%s", marathon, uri)
 
 	var jsonBody []byte
@@ -237,18 +237,11 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 		}
 	}
 
-	// Make the http request to Marathon
-	request, err := http.NewRequest(method, url, bytes.NewReader(jsonBody))
+	// step: create an API request
+	request, err := r.apiRequest(method, url, bytes.NewReader(jsonBody))
 	if err != nil {
 		return err
 	}
-
-	// Add any basic auth and the content headers
-	if r.config.HTTPBasicAuthUser != "" {
-		request.SetBasicAuth(r.config.HTTPBasicAuthUser, r.config.HTTPBasicPassword)
-	}
-	request.Header.Add("Content-Type", "application/json")
-	request.Header.Add("Accept", "application/json")
 
 	response, err := r.httpClient.Do(request)
 	if err != nil {
@@ -278,6 +271,25 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 	}
 
 	return NewAPIError(response.StatusCode, respBody)
+}
+
+// apiRequest creates a default API request
+func (r *marathonClient) apiRequest(method, url string, reader io.Reader) (*http.Request, error) {
+	// Make the http request to Marathon
+	request, err := http.NewRequest(method, url, reader)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add any basic auth and the content headers
+	if r.config.HTTPBasicAuthUser != "" && r.config.HTTPBasicPassword != "" {
+		request.SetBasicAuth(r.config.HTTPBasicAuthUser, r.config.HTTPBasicPassword)
+	}
+
+	request.Header.Add("Content-Type", "application/json")
+	request.Header.Add("Accept", "application/json")
+
+	return request, nil
 }
 
 var oneLogLineRegex = regexp.MustCompile(`(?m)^\s*`)
--- a/client_test.go
+++ b/client_test.go
@@ -38,6 +38,78 @@ func TestGetMarathonURL(t *testing.T) {
 	assert.Equal(t, endpoint.Client.GetMarathonURL(), endpoint.URL)
 }
 
+func TestAPIRequest(t *testing.T) {
+	cases := []struct {
+		Username       string
+		Password       string
+		ServerUsername string
+		ServerPassword string
+		Ok             bool
+	}{
+		{
+			Username:       "should_pass",
+			Password:       "",
+			ServerUsername: "",
+			ServerPassword: "",
+			Ok:             true,
+		},
+		{
+			Username:       "bad_username",
+			Password:       "",
+			ServerUsername: "test",
+			ServerPassword: "password",
+			Ok:             false,
+		},
+		{
+			Username:       "test",
+			Password:       "bad_password",
+			ServerUsername: "test",
+			ServerPassword: "password",
+			Ok:             false,
+		},
+		{
+			Username:       "",
+			Password:       "",
+			ServerUsername: "test",
+			ServerPassword: "password",
+			Ok:             false,
+		},
+		{
+			Username:       "test",
+			Password:       "password",
+			ServerUsername: "test",
+			ServerPassword: "password",
+			Ok:             true,
+		},
+	}
+	for i, x := range cases {
+		var endpoint *endpoint
+
+		config := NewDefaultConfig()
+		config.HTTPBasicAuthUser = x.Username
+		config.HTTPBasicPassword = x.Password
+
+		endpoint = newFakeMarathonEndpoint(t, &configContainer{
+			client: &config,
+			server: &serverConfig{
+				username: x.ServerUsername,
+				password: x.ServerPassword,
+			},
+		})
+
+		_, err := endpoint.Client.Applications(nil)
+
+		if x.Ok && err != nil {
+			t.Errorf("case %d, did not expect an error: %s", i, err)
+		}
+		if !x.Ok && err == nil {
+			t.Errorf("case %d, expected to received an error", i)
+		}
+
+		endpoint.Close()
+	}
+}
+
 func TestOneLogLine(t *testing.T) {
 	in := `
 	a
--- a/subscription.go
+++ b/subscription.go
@@ -86,7 +86,7 @@ func (r *marathonClient) SubscriptionURL() string {
 	return fmt.Sprintf("http://%s:%d%s", r.ipAddress, r.config.EventsPort, defaultEventsURL)
 }
 
-// RegisterSubscription registers ourselves with Marathon to receive events from configured transport facility
+// registerSubscription registers ourselves with Marathon to receive events from configured transport facility
 func (r *marathonClient) registerSubscription() error {
 	switch r.config.EventsTransport {
 	case EventsTransportCallback:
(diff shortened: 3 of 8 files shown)


gambol99:
fixes https://github.com/gambol99/go-marathon/issues/148

gambol99:
@timoreimann

subscription.go
```diff
@@ -157,16 +157,23 @@ func (r *marathonClient) registerSSESubscription() error {
...
-	stream, err := eventsource.Subscribe(url, "")
+	// step: create a new request for the stream
+	request, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return err
+	}
+	// step: add any basic authentication
+	if r.config.HTTPBasicAuthUser != "" && r.config.HTTPBasicPassword != "" {
+		request.SetBasicAuth(r.config.HTTPBasicAuthUser, r.config.HTTPBasicPassword)
+	}
```
timoreimann: We already run a similar, yet already slightly deviating logic in [apiCall()](https://github.com/gambol99/go-marathon/blob/fbaf44af7084997a4bb1fd1749a271640d17f671/client.go#L247). What would you think about extracting it into a central utility function that we could use from both call sites?

timoreimann:
I think we should cover the new code by some test. In #142, @dmajere [suggested](https://github.com/gambol99/go-marathon/pull/142#issuecomment-221016520) to inject something we can mock/stub. Another alternative would be to run against our existing YAML-based fake and see if we can define some expectation on the existence of the required headers.

timoreimann:
Review done. :-)

gambol99:
@timoreimann .. not sure if this is best way of doing it .. so open to opinions. I've wrapped the auth under and common method and added a mock basic auth middleware to the test server

client.go
```diff
@@ -221,35 +222,28 @@ func (r *marathonClient) apiDelete(uri string, post, result interface{}) error {
...
 	if body != nil {
 		jsonBody, err = json.Marshal(body)
 		if err != nil {
 			return err
 		}
 	}
 
-	// Make the http request to Marathon
-	request, err := http.NewRequest(method, url, bytes.NewReader(jsonBody))
+	// step: create an api request
```
timoreimann: nit-pick: `s/api/API/`.

client.go
```diff
@@ -285,6 +279,25 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 	return apiErr
 }
 
+// apiRequest creates a default api request
```
timoreimann: nit-pick: `s/default api request/default API request/`.

client.go
```diff
@@ -285,6 +279,25 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
...
+func (r *marathonClient) apiRequest(method, url string, reader io.Reader) (*http.Request, error) {
+	// Make the http request to Marathon
```
timoreimann: "Make" could be misinterpreted as in "send" whereas it means "create" here. Maybe use the latter instead?

client_test.go
```diff
@@ -38,6 +38,53 @@ func TestGetMarathonURL(t *testing.T) {
 	assert.Equal(t, endpoint.Client.GetMarathonURL(), endpoint.URL)
 }
 
+func TestApiRequest(t *testing.T) {
```
timoreimann: `s/Api/API/` to satisfy [Go's convention on initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

client_test.go
```diff
@@ -38,6 +38,53 @@ func TestGetMarathonURL(t *testing.T) {
...
+	cases := []struct {
+		Auth     bool
+		Username string
+		Password string
+		Ok       bool
+	}{
+		{
+			Username: "bad",
```
timoreimann: Does it make sense to set `Username` if `Auth` is false at the same time? The test seems to ignore credentials if we disable `Auth`.

client_test.go
```diff
@@ -38,6 +38,53 @@ func TestGetMarathonURL(t *testing.T) {
...
+			Ok:       true,
+		},
+		{
+			Auth:     true,
+			Username: "bad",
+		},
```
timoreimann: I'd also define `Password` and `Ok` explicitly here in order to improve readability of the test.

client_test.go
```diff
@@ -38,6 +38,53 @@ func TestGetMarathonURL(t *testing.T) {
...
+		{
+			Auth:     true,
+			Username: "test",
+			Password: "test",
+			Ok:       true,
+		},
+	}
+	for i, x := range cases {
+		var endpoint *endpoint
+		switch x.Auth {
```
timoreimann: We could get rid of `Auth` and the switch statement altogether with the following adjustments:

- Always pass username and password to the configuration and `newFakeAuthMarathonEndpoint`.
- Change `basicAuthMiddleware` to short-circuit / ignore authentication if either username and password are empty.

Note that the latter actually reflects [Marathon's real-world behavior](https://github.com/gambol99/go-marathon/pull/142#issuecomment-221151346): One must provide both a non-empty username and password in the server configuration, or otherwise end up with Marathon not doing any authentication at all.

testing_utils_test.go
```diff
@@ -79,22 +78,28 @@ func getTestURL(urlString string) string {
 	return fmt.Sprintf("%s://%s", parsedURL.Scheme, strings.Join([]string{parsedURL.Host, parsedURL.Host, parsedURL.Host}, ","))
 }
 
+func newFakeAuthMarathonEndpoint(t *testing.T, config *Config, username, password string) *endpoint {
```
timoreimann: With the very recent merge of #158, I think we should take advantage of the new `ServerConfig` type and amend it by `username` and `password` fields. That way, we can reuse the existing method calls and don't have to define `newFakeAuthMarathonEndpoint`.

testing_utils_test.go
```diff
@@ -142,6 +145,24 @@ func newFakeMarathonEndpoint(t *testing.T, config *Config) *endpoint {
...
+	unauthorized := "{\"message\": \"invalid username or password\"}"
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		u, p, found := r.BasicAuth()
+		if !found && (username != "" || password != "") {
+			http.Error(w, unauthorized, 401)
+			return
+		}
+		if username != u || password != p {
+			http.Error(w, unauthorized, 401)
```
timoreimann: FWIW, we would also run into this case if the username and password passed to `basicAuthMiddleware` were both empty but at least one of the credentials parameters contained in the request was not. As described in a comment above, that's actually not how Marathon behaves: It skips authentication if either username and password is empty.

Short-circuiting in the case were either is empty would solve this.

testing_utils_test.go
```diff
@@ -142,6 +145,24 @@ func newFakeMarathonEndpoint(t *testing.T, config *Config) *endpoint {
...
+			return
+		}
```
timoreimann: I had to run through the various possible code paths a couple of times in order to convince myself that this is probably correct. This makes me think that the middleware might be complex enough to deserve a test of its own.

FWIW, at my organization if we need to write a test for a test helper located in a `*_test.go` file, we put it in a `*_test_test.go` file. Looks a bit weird but actually follows the naming logic quite well from my perspective.

timoreimann:
I like the approach in general. Left some comments to a few details.

Note that your build is conflicting with master, most likely due to the changes from #158 merged in earlier.

gambol99:
hi @timoreimann  ... i've made the various changes ... you mind taking a look :-)

client.go
```diff
@@ -221,35 +222,28 @@ func (r *marathonClient) apiDelete(uri string, post, result interface{}) error {
 }
 
 func (r *marathonClient) apiCall(method, uri string, body, result interface{}) error {
+	var jsonBody []byte
```
timoreimann: FWIW, they say you should declare variables as close to their first usage as possible. :-)

testing_utils_test.go
```diff
@@ -40,9 +40,14 @@ const (
...
 	fakeAPIFilename   = "./tests/rest-api/methods.yml"
-	fakeAPIPort       = 3000
 )
 
+var (
+	fakeResponses map[string]string
+	once          sync.Once
+)
+
+// restMethod represents a fake response
```
timoreimann: maybe better: `restMethod represents an expected HTTP method and an associated fake response`

testing_utils_test.go
```diff
@@ -54,6 +59,20 @@ type restMethod struct {
...
 }
 
+// serverConfig holds the Marathon server configuration
+type serverConfig struct {
+	// The version of Marathon
+	version string
+}
+
+// configContainer holds both server and client Marathon configuration
+type configContainer struct {
```
timoreimann: Appreciate the move of this struct to the test namespace. :+1:

testing_utils_test.go
```diff
@@ -54,6 +59,20 @@ type restMethod struct {
...
+	client   *Config
+	server   *serverConfig
+	username string
+	password string
```
timoreimann: Unless I'm mistaken, username and password are only used on the (fake) server side. They should thus live in `config.ServerConfig`.

testing_utils_test.go
```diff
@@ -73,83 +92,51 @@ type fakeEvent struct {
...
-				key += fmt.Sprintf(":%s", method.Version)
-			}
-			uris[key] = &method.Content
-		}
-	})
+func newFakeMarathonEndpoint(t *testing.T, configs *configContainer) *endpoint {
+	// step: read in the fake responses if required
+	newFakeMarathonResponses(t)
 
+	// step: create a fake sse event service
```
timoreimann: nit-pick: `s/sse/SSE/`.

testing_utils_test.go
```diff
@@ -73,83 +92,51 @@ type fakeEvent struct {
...
+			http.Error(writer, `{"message": "not found"}`, 404)
 			return
 		}
-		http.Error(writer, `{"message": "not found"}`, 404)
-	})
+		writer.Header().Add("Content-Type", "application/json")
+		writer.Write([]byte(content))
+	}))
 
+	// step: create http test server
```
timoreimann: nit-pick: `s/http/HTTP/`.

testing_utils_test.go
```diff
@@ -165,6 +152,52 @@ func newFakeMarathonEndpoint(t *testing.T, configs *ConfigContainer) *endpoint {
 	}
 }
 
+// basicAuthMiddleware handles basic auth
+func basicAuthMiddleware(config *configContainer, next http.HandlerFunc) func(http.ResponseWriter, *http.Request) {
```
timoreimann: It should suffice to pass the username and password parameters instead of the entire configuration to this function as we don't need the rest.

testing_utils_test.go
```diff
@@ -165,6 +152,52 @@ func newFakeMarathonEndpoint(t *testing.T, configs *ConfigContainer) *endpoint {
...
+	unauthorized := "{\"message\": \"invalid username or password\"}"
```
timoreimann: We can get rid of the escaped quotes and increase readability by wrapping the string in back ticks (`` ` ``).

testing_utils_test.go
```diff
@@ -165,6 +152,52 @@ func newFakeMarathonEndpoint(t *testing.T, configs *ConfigContainer) *endpoint {
...
+				return
+			}
+		}
+
+		next(w, r)
+	}
+}
+
+// newFakeMarathonResponses reads in the marathon fake responses from the yaml file
+func newFakeMarathonResponses(t *testing.T) {
```
timoreimann: The `new` prefix is conventionally reserved for functions returning a pointer to a type. I think we should call the method `initFakeMarathonResponses ` here.

Personally I think it'd be nicer to have the function not take a `*testing.T` parameter and instead return an `error`. Helper functions shouldn't decide when a test must fail.

timoreimann:
Looking pretty good now. Left a few more notes, mostly detail-ish. :-)

timoreimann:
I believe this ticket also

Closes #142 

once it's merged.

client.go
gambol99: oh ... i always preferred the C approach and chunk everything at the top :-) ... I'll stick to the default though!

gambol99:
@timoreimann ... changes made .. cheers for the review :-)

testing_utils_test.go
gambol99: problem with not passing the *testing.T is Once.Do() doesn't return an error ...

timoreimann:
Awesome work, LGTM!

Let's 🚢 !

PR #157, opened 2016-06-05
Fix internal golint errors

timoreimann:
This PR fixes all remaining, _internal_ golint errors.

By internal, we mean those that are subject to go-marathon itself and not imposed by a third-party package imported.


timoreimann:
@gambol99 should be a quick one. :-)

gambol99:
LGTM

Issue #153, opened 2016-05-27
go-marathon does not work with marathon 1.1.1 due to changed deployments/steps schema


kesselborn:
go-marathon does not work correctly anymore with marathon 1.1.1 as the deployments/steps schema was changed.
Up to now it looks more like this is a [marathon bug](https://github.com/mesosphere/marathon/issues/3820) which currently is set to fixed but does not fix the real problem (see [comment on the issue](https://github.com/mesosphere/marathon/issues/3820#issuecomment-222201282) ).
This issue is to track the solution and create a pull request should the new schema be correct. Happy to provide a pull request once the situation is more clear.

timoreimann:
PR would be very much appreciated once the situation has cleared up. :+1:

timoreimann:
(updated the first broken link in the description. Github issue references don't seem to work when used together with custom links)

kesselborn:
@timoreimann oh, thanks

timoreimann:
@kesselborn with #158 being merged, do you think we should  keep this issue open in order to track how mesosphere/marathon#3965 evolves? Or close this one instead and open a new issue if the API changes again?

timoreimann:
Closing now. We can easily open another issue should the API change again.

PR #158, opened 2016-06-06
Fix marshalling of DeploymentStep for Marathon v1.X API

kevinschoon:
Hi -- 

I wrote a quick fix for https://github.com/gambol99/go-marathon/issues/153. It's not necessarily ideal but hopefully should suffice until https://github.com/mesosphere/marathon/issues/3820 is truly resolved. Also hoping to provide a fix for https://github.com/Banno/terraform-provider-marathon/issues/25.

Thanks!


deployment.go
```diff
@@ -28,7 +29,8 @@ type Deployment struct {
 	CurrentStep    int                 `json:"currentStep"`
 	TotalSteps     int                 `json:"totalSteps"`
 	AffectedApps   []string            `json:"affectedApps"`
-	Steps          [][]*DeploymentStep `json:"steps"`
+	Steps          [][]*DeploymentStep `json:"-"`
+	XXStepsRaw     json.RawMessage     `json:"steps"`
```
timoreimann: Maybe the `XX` notation is conventional and I simply don't know about it, but I'd still love to see a comment next to this field saying that it is exported for internal purposes only.

deployment.go
```diff
@@ -72,7 +74,38 @@ func (r *marathonClient) Deployments() ([]*Deployment, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	// Allows loading of deployment steps from the Marathon v1.X API
+	// Implemented as a Quick fix for issue https://github.com/gambol99/go-marathon/issues/153
+	if r.config.V1X {
```
timoreimann: I'd rather not have this be a switch you need to enable first. Maybe it's possible to unmarshal both pre-1.1 and 1.1 versions and pick the one we find populated? This would also allow for a smoother transition on the consumer end if upstream decides to go back to the pre-1.1 format.

deployment.go
```diff
@@ -72,7 +74,38 @@ func (r *marathonClient) Deployments() ([]*Deployment, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	// Allows loading of deployment steps from the Marathon v1.X API
+	// Implemented as a Quick fix for issue https://github.com/gambol99/go-marathon/issues/153
```
timoreimann: nit-pick: I'd drop marking this as a quick fix. Those tend to outlive their purpose frequently. :-)

timoreimann:
The change should definitely be accompanied by a test. Should be fairly easy though to extend our YAML definition accordingly.

timoreimann:
I had hoped for Mesosphere to react more quickly on the upstream issue so that we don't need this. Unfortunately, it looks like that's not happening, so support for the new (intermediary?) 1.1 format looks desirable.

Thanks for the PR -- I left a few remarks.

kevinschoon:
Appreciate the feedback. 

Hopefully I addressed all the items you mentioned in my last commit. 

I changed a little bit of the logic in your test suite to support returning different responses based on an explicit version that can be passed in the yaml method definition file.

Kevin

config.go
```diff
@@ -45,6 +45,8 @@ type Config struct {
 	LogOutput io.Writer
 	// HTTPClient is the http client
 	HTTPClient *http.Client
+	// Specify an Explicit Marathon Version
+	Version string
```
timoreimann: `marathon.Config` is meant to hold client-side options. `Version` is only relevant for the (fake) server end, however. It could mislead client users into thinking that they can make use the field somehow.

My suggestion would be to change [newFakeMarathonEndpoint](https://github.com/kevinschoon/go-marathon/blob/2bdb699c4c6b14c5b58a92879c9fc626d5dec157/testing_utils_test.go#L84)'s signature such that it accepts a (newly defined) configuration container struct holding both client-side and server-side configurations. The former would be what we currently pass into the function; the latter would contain just the version specifier. So the updated signature could look similar to

`func newFakeMarathonEndpoint(t *testing.T, configs *ConfigContainer) *endpoint`

with `ConfigContainer` being something like

```
type ConfigContainer struct {
  client *config.Config
  server *config.ServerConfig
}
```

Since we currently only have a handful of cases where we actually pass a non-nil configuration to `newFakeMarathonEndpoint`, we wouldn't even need to change a lot of places to accommodate for the new prototype definition.

WDYT?

deployment.go
```diff
@@ -44,6 +46,14 @@ type DeploymentStep struct {
 	App    string `json:"app"`
 }
 
+// DeploymentStepV1 is a Marathon v1.X-compatible DeploymentStep
+type DeploymentStepV1 struct {
+	Actions []struct {
+		Type string `json:"type"`
```
timoreimann: I believe the string type and JSON tag should be called `Action` and `action`, respectively: This is what [upstream changed it to](https://github.com/mesosphere/marathon/commit/595eba7d86dd59bb85ee30c03ffd3481136c4dcf). See also @kesselborn's [comparison on the upstream issue](https://github.com/mesosphere/marathon/issues/3820#issuecomment-222201282).

Tests would need to be updated as well.

deployment.go
```diff
@@ -72,7 +82,28 @@ func (r *marathonClient) Deployments() ([]*Deployment, error) {
...
 		return nil, err
 	}
-
+	// Allows loading of deployment steps from the Marathon v1.X API
+	// Implements a fix for issue https://github.com/gambol99/go-marathon/issues/153
+	for _, deployment := range deployments {
+		// Unmarshal pre-v1.X step
+		if err := json.Unmarshal(deployment.XXStepsRaw, &deployment.Steps); err != nil {
+			deployment.Steps = make([][]*DeploymentStep, 0)
+			steps := make([]*DeploymentStepV1, 0)
```
timoreimann: nit-pick: You can also use `var steps []*DeploymentStepV1` here to save one allocation.

deployment_test.go
```diff
@@ -37,6 +37,23 @@ func TestDeployments(t *testing.T) {
 	assert.Equal(t, len(deployment.Steps), 1)
 }
 
+func TestDeploymentsV1(t *testing.T) {
+	cfg := NewDefaultConfig()
+	cfg.Version = "v1.1.1"
```
timoreimann: nit-pick: I'd remove the `v` prefix. It should be clear that field contains numeric version specifiers only.

timoreimann:
Placed some more comments. I really like the current implementation approach, including the extension to the test suite supporting version specifiers. 👍

deployment.go
kevinschoon: I had seen that however that change does not exist in the current release (v1.1.1) which I am attempting to fix and instead be included with the next [release](https://github.com/mesosphere/marathon/milestones/0.16.0). I feel it is best to wait until that version is released, and then update since this PR aims to explicitly support **v1.1.1**.

config.go
kevinschoon: I initially was planning to change the signature and implement something like what you're describing but ultimately wanted to avoid updating all of the test code. Although there is currently no other use for specifying the version in the main configuration I could potentially see other parts of the library benefiting from knowing the version of Marathon. It could even be populated programatically when instantiating the client by making a call to the `/v2/info` endpoint.

kesselborn:
I created a new bug on marathon as it's not possible to re-open the old one. Nothing against this implementation, but I am not sure this is the best solution to fix this problem (although the most pragmatic one).

config.go
timoreimann: Teaching go-marathon to understand the Marathon version it's talking to (either manually or automatically) in a reliable way introduces a certain layer of complexity with a number of corner cases that we should consider carefully. This probably deserves a separate issue and discussion, so I'd rather avoid extending the consumer API surface along this PR by something we don't have a clear usage pattern for yet.

I believe the amount of test code we'd need to update in order for the signature change to happen should be fairly manageable: We have 4 or 5 call sites (including the new one from this PR) where the existing `config` parameter is not nil; those would need to be modified to pass a new configuration container struct instead. The internal change to `newFakeMarathonEndpoint` should be very simple too. If you prefer to minimize the overall effort, we could also create a new function `newFakeMarathonEndpointFromConfigs(Config, ServerConfig)` that you can call directly from the new test and the existing `newFakeMarathonEndpoint` (with the `ServerConfig` argument being nil for the latter case). A follow-up PR could then clean up all remaining call sites.

deployment.go
timoreimann: Oh yes, you're absolutely right. Missed to see that the latest change hasn't been released yet. 👍

timoreimann:
@kesselborn I think we both agree that the best solution is to have upstream fix the problem (either by reverting to the old format or declaring a breaking change along with a proper minor version bump). Feel free to share any other concerns you see with the PR at hand.

config.go
kevinschoon: I went with your initial suggestion in my latest commit.

kevinschoon:
I added the changes suggested by @timoreimann in my last commits. Ultimately I agree with @kesselborn that an upstream fix by Mesosphere is ideal and the best solution. This branch fixes the immediate bug I was trying to circumvent: https://github.com/Banno/terraform-provider-marathon/issues/25

timoreimann:
LGTM.

Thanks a lot!

kesselborn:
@kevinschoon @timoreimann sorry that I wasn't really helpful here and instead only raised my concerns (and then did not reply again) ... I did not find the time. Thanks a lot for the work you did.

PR #161, opened 2016-06-07
Support marathon dcos token

jacob-koren:


client.go
```diff
@@ -30,6 +30,8 @@ import (
 	"time"
 )
 
+var MARATHON_TOKEN = ""
```
timoreimann: We shouldn't work with globals here -- they pose a risk for concurrent usage. A better place is likely our [configuration](https://github.com/gambol99/go-marathon/blob/697a616b3996092969b86950b1e555e3b4391945/config.go#L29).

client.go
```diff
@@ -168,6 +170,8 @@ type marathonClient struct {
 //		config:			the configuration to use
 func NewClient(config Config) (Marathon, error) {
 	// step: if no http client, set to default
+
+	fmt.Println(config.URL)
```
timoreimann: In general, logging should happen through the [log writer](https://github.com/gambol99/go-marathon/blob/697a616b3996092969b86950b1e555e3b4391945/config.go#L45). I don't really see a need to log the URL though. Could you please explain why you think it's necessary?

client.go
```diff
@@ -204,6 +208,7 @@ func (r *marathonClient) Ping() (bool, error) {
 	return true, nil
 }
 
+
```
timoreimann: Please remove the extra blank line.

client.go
```diff
@@ -220,14 +225,15 @@ func (r *marathonClient) apiDelete(uri string, post, result interface{}) error {
 	return r.apiCall("DELETE", uri, post, result)
 }
 
+
```
timoreimann: Please remove the extra blank line.

client.go
```diff
@@ -220,14 +225,15 @@ func (r *marathonClient) apiDelete(uri string, post, result interface{}) error {
...
 func (r *marathonClient) apiCall(method, uri string, body, result interface{}) error {
 	// Get a member from the cluster
 	marathon, err := r.cluster.GetMember()
 	if err != nil {
 		return err
 	}
 
-	url := fmt.Sprintf("%s/%s", marathon, uri)
+	url := fmt.Sprintf("%s/%s", marathon + "/marathon", uri)
```
timoreimann: Where does this endpoint come from? Is that possibly DCOS-specific? Whatever it does, it's breaking the tests right now.

timoreimann:
Would you mind summarizing the purpose of your change in the issue description? I for one am not too familiar with DCOS and the token you're referring to. A link to some piece of documentation would be great too.

We should also consider adding a test for the functionality you're adding.

Thanks!

jacob-koren:
Hi, you are absolutely right. I'll issue a new pull request with the changes.
DCOS is using marathon (by default) as the container scheduler of the system. In order to perform any operation against a marathon that resides in this environment (or any other service) you must add a DCOS token.

timoreimann:
Thanks for the explanation. Is that also why you require a different end point for the Marathon API?

You don't need to create a new PR; simply push additional changes to your existing branch until we're good to go.

jacob-koren:
Yes. Because the root URL in dcos cluster is the cluster itself, marathon is a service that runs inside the cluster in /marathon path.

jacob-koren:
Btw, you are missing the new optional parameters in marathon ports mapping (labels and name): 


      {
        "containerPort": 8000,
        "hostPort": 0,
        "servicePort": 10005,
        "protocol": "tcp",
        **"name": "debug-port",
        "labels": {}**
}

jacob-koren:
Hi, are we ready to merge?

client.go
```diff
@@ -250,6 +256,10 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 	request.Header.Add("Content-Type", "application/json")
 	request.Header.Add("Accept", "application/json")
 
+	if r.config.MarathonToken != "" {
+		request.Header.Add("Authorization", "token=" + r.config.MarathonToken)
```
timoreimann: This will overwrite any basic authorization set before. While I suppose it doesn't make sense to set both the token and basic auth credentials at the same time, we should leave a comment next to the token field about this circumstance and also have a test for this.

client.go
```diff
@@ -286,4 +296,4 @@ var oneLogLineRegex = regexp.MustCompile(`(?m)^\s*`)
 // escapes new line characters.
 func oneLogLine(in []byte) []byte {
 	return bytes.Replace(oneLogLineRegex.ReplaceAll(in, nil), []byte("\n"), []byte("\\n "), -1)
-}
+}
```
timoreimann: Missing new line?

config.go
```diff
@@ -35,12 +35,16 @@ type Config struct {
 	EventsPort int
 	// the interface we should be listening on for events
 	EventsInterface string
+	// dcos token
+	MarathonToken string
```
timoreimann: You have `MarathonToken` defined twice. Please remove this instance.

config.go
```diff
@@ -35,12 +35,16 @@ type Config struct {
...
 	// HTTPBasicAuthUser is the http basic auth
 	HTTPBasicAuthUser string
 	// HTTPBasicPassword is the http basic password
 	HTTPBasicPassword string
 	// CallbackURL custom callback url
 	CallbackURL string
+	// Token for DCOS environment
```
timoreimann: We should mention here that the token will overwrite any basic auth credentials.

config.go
```diff
@@ -35,12 +35,16 @@ type Config struct {
...
+	MarathonToken string
```
timoreimann: I'm not sure about the name. Is the token specific to DCOS's "flavor" of Marathon or something that vanilla Marathon supports too? Can you possibly point me at some piece of documentation where the token is described?

timoreimann:
Even if we wanted to merge, we couldn't because your branch conflicts with master. Please rebase and fix any conflicts.

Please also add a test. You can [define a new test method](https://github.com/gambol99/go-marathon/blob/be8af73c740347e1740996296baef1000733b3d9/tests/rest-api/methods.yml) for the fake marathon server to run against.

config.go
jacob-koren: Maybe i should call it DCOS token: https://dcos.io/docs/1.7/administration/security/auth-api/

timoreimann: Sounds good to me. :+1:

jacob-koren:
Fixed according to your comments. Please review.

client.go
```diff
@@ -243,6 +251,17 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 		return err
 	}
 
+	// Add any basic auth and the content headers
```
gambol99: Can you remove this section and update [apiRequest()](https://github.com/gambol99/go-marathon/blob/master/client.go#L277) instead ... Authorization was moved into a common method due to issues with marathon sse events

config.go
```diff
@@ -41,6 +41,8 @@ type Config struct {
 	HTTPBasicPassword string
 	// CallbackURL custom callback url
 	CallbackURL string
+	// Token for DCOS environment, This will override the Authorization header
```
gambol99: s/Token/DOCSToken/

testing_utils_test.go
```diff
@@ -119,8 +121,8 @@ func newFakeMarathonEndpoint(t *testing.T, configs *configContainer) *endpoint {
 
 	// step: create the HTTP router
 	mux := http.NewServeMux()
-	mux.HandleFunc("/v2/events", basicAuthMiddleware(configs.server, eventSrv.Handler("event")))
-	mux.HandleFunc("/", basicAuthMiddleware(configs.server, func(writer http.ResponseWriter, reader *http.Request) {
+	mux.HandleFunc("/v2/events", AuthMiddleware(configs.server, eventSrv.Handler("event")))
```
gambol99: s/AuthMiddleware/authMiddleware/

testing_utils_test.go
```diff
@@ -177,6 +179,46 @@ func basicAuthMiddleware(server *serverConfig, next http.HandlerFunc) func(http.
 	}
 }
 
+// basicAuthMiddleware handles basic auth
+func AuthMiddleware(server *serverConfig, next http.HandlerFunc) func(http.ResponseWriter, *http.Request) {
```
gambol99: Same as above ... simply lowercase and update the method comment

testing_utils_test.go
```diff
@@ -177,6 +179,46 @@ func basicAuthMiddleware(server *serverConfig, next http.HandlerFunc) func(http.
...
+			headerValue := r.Header.Get("Authorization")
+			// step: if no auth found, error it
+			if headerValue == "" {
+				http.Error(w, unauthorized, 401)
+				return
+			}
+
+			s := strings.Split(headerValue, "=")
+
+			fmt.Println("serverToken is: " + server.dcosToken + " and request token is: " + s[1])
```
gambol99: can we remove debug line ..

client.go
```diff
@@ -222,12 +222,20 @@ func (r *marathonClient) apiDelete(uri string, post, result interface{}) error {
...
 	// Get a member from the cluster
 	marathon, err := r.cluster.GetMember()
 	if err != nil {
 		return err
 	}
-	url := fmt.Sprintf("%s/%s", marathon, uri)
+
+	var url string
+
+	if r.config.DCOSToken != "" {
```
gambol99: Perhaps in another PR we can change this logic and place into Config itself ... i.e 
```GO
marathonURL := "http://10.241.1.71:8080/marathon,http://10.241.1.81:8080/marathon"
config := marathon.NewDefaultConfig()
config.URL = marathonURL
```
I think at the moment, the GetMember() will only return scheme://hostname ... I'll chunk it into an issue

config.go
jacob-koren: Can you elaborate?

jacob-koren: Nevermind, got it.

testing_utils_test.go
jacob-koren: Wierd, thought i removed it.

gambol99:
lgtm ... @timoreimann? ...

client_test.go
```diff
@@ -124,3 +124,71 @@ func TestOneLogLine(t *testing.T) {
...
 }
+
+func TestAPIRequestDCOS(t *testing.T) {
+	cases := []struct {
+		DCOSToken       string
+		ServerDCOSToken string
+		Username        string
+		Password        string
+		ServerUsername  string
+		ServerPassword  string
```
timoreimann: We don't seem to use the username and password fields in any of the tests. I think we should just drop them from the cases struct and, subsequently, not include them into the configuration struct (explicitly).

timoreimann:
Looking pretty sound to me already.

I had some time to go over the upstream documentation of the feature. AFAICS, the `ipAddress` object itself and any of its nested fields can (and partially must) be set by the client. Due to the way JSON serialization in Go works and the fact that Marathon only offers PATCH semantics through the PUT verb, we need to pointerize everything so that clients have the possibility to clear settings. This already holds for some but not all fields.

Along with that, we should also have a set of convenience setters. Could you please take a look at how we do this with other fields (such as (top-level) labels) and mimic things for the new structs? We usually just provide a pair of `Empty*` and `Set*` methods. Since `IPAddressPerTask` is hierarchical, I believe it lends itself to defining the setters at the appropriate levels, respectively. For instance, the setter for `IPAddressPerTask` should be attached to `Application` while those for the field's sub-fields (i.e., `Groups`, `Labels`, and `Discovery`) should be hosted at `IPAddressPerTask`.

gambol99:
LTGM ... merging ...

"I had some time to go over the upstream documentation of the feature. AFAICS, the ipAddress object itself and any of its nested fields can" .. @timoreimann, this is related to PR #163 no?

timoreimann:
@gambol99 Oh yes, absolutely! Thanks for pointing this out. Will move the comment over to the other PR.

PR #163, opened 2016-06-08
Added IpAddress for applications (Ip Per Task)

tommyvicananza:
IpAddress is used by Ip Per Task Funcionality https://mesosphere.github.io/marathon/docs/ip-per-task.html

### DIFF
--- a/application.go
+++ b/application.go
@@ -34,6 +34,25 @@ type Applications struct {
 	Apps []Application `json:"apps"`
 }
 
+// IpAddress is used by Ip Per Task Funcionality https://mesosphere.github.io/marathon/docs/ip-per-task.html
+type IpAddress struct {
+	Groups []string `json:"groups,omitempty"`
+	Labels map[string]string `json:"labels,omitempty"`
+	Discovery *Discovery `json:"discovery,omitempty"`
+}
+
+// Discovery provides info about ports expose by Ip Per Task Funcionality
+type Discovery struct {
+	Ports []*Ports `json:"ports,ommitempty"`
+}
+
+// Ports provides info about ports used by Ip Per Task
+type Ports struct {
+	Number int `json:"number,omitempty"`
+	Name string `json:"name:omitempty"`
+	Labels map[string]string
+}
+
 // Application is the definition for an application in marathon
 type Application struct {
 	ID                    string              `json:"id,omitempty"`
@@ -69,6 +88,7 @@ type Application struct {
 	AcceptedResourceRoles []string            `json:"acceptedResourceRoles,omitempty"`
 	LastTaskFailure       *LastTaskFailure    `json:"lastTaskFailure,omitempty"`
 	Fetch                 []Fetch             `json:"fetch"`
+	IpAddress							*IpAddress					`json:"ipAddress,omitempty"`
 }
 
 // ApplicationVersions is a collection of application versions for a specific app in marathon
--- a/application.go
+++ b/application.go
@@ -34,19 +34,19 @@ type Applications struct {
 	Apps []Application `json:"apps"`
 }
 
-// IpAddress is used by Ip Per Task Funcionality https://mesosphere.github.io/marathon/docs/ip-per-task.html
-type IpAddress struct {
+// IPAddressPerTask is used by IP-per-task functionality Funcionality https://mesosphere.github.io/marathon/docs/ip-per-task.html
+type IPAddressPerTask struct {
 	Groups []string `json:"groups,omitempty"`
 	Labels map[string]string `json:"labels,omitempty"`
 	Discovery *Discovery `json:"discovery,omitempty"`
 }
 
-// Discovery provides info about ports expose by Ip Per Task Funcionality
+// Discovery provides info about ports expose by Ip-per-task Funcionality
 type Discovery struct {
 	Ports []*Ports `json:"ports,ommitempty"`
 }
 
-// Ports provides info about ports used by Ip Per Task
+// Ports provides info about ports used by Ip-per-task
 type Ports struct {
 	Number int `json:"number,omitempty"`
 	Name string `json:"name:omitempty"`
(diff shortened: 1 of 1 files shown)


application.go
```diff
@@ -34,6 +34,25 @@ type Applications struct {
 	Apps []Application `json:"apps"`
 }
 
+// IpAddress is used by Ip Per Task Funcionality https://mesosphere.github.io/marathon/docs/ip-per-task.html
```
timoreimann: Please call the struct `IPAddress` to satisfy [Golang's initialism rule](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

Style improvements and typo fixes:

_Ip Per Task Funcionality_ ==> _IP-per-task functionality_

application.go
```diff
@@ -34,6 +34,25 @@ type Applications struct {
...
+type IpAddress struct {
```
timoreimann: `IpAddress` -> `IPAddress`

application.go
```diff
@@ -34,6 +34,25 @@ type Applications struct {
...
+	Groups []string `json:"groups,omitempty"`
+	Labels map[string]string `json:"labels,omitempty"`
+	Discovery *Discovery `json:"discovery,omitempty"`
+}
+
+// Discovery provides info about ports expose by Ip Per Task Funcionality
```
timoreimann: _Ip Per Task Funcionality_ ==> _IP-per-task functionality_

application.go
```diff
@@ -34,6 +34,25 @@ type Applications struct {
...
+type Discovery struct {
+	Ports []*Ports `json:"ports,ommitempty"`
+}
+
+// Ports provides info about ports used by Ip Per Task
```
timoreimann: _Ip Per Task_ ==> _IP-per-task_

application.go
```diff
@@ -69,6 +88,7 @@ type Application struct {
 	AcceptedResourceRoles []string            `json:"acceptedResourceRoles,omitempty"`
 	LastTaskFailure       *LastTaskFailure    `json:"lastTaskFailure,omitempty"`
 	Fetch                 []Fetch             `json:"fetch"`
+	IpAddress							*IpAddress					`json:"ipAddress,omitempty"`
```
timoreimann: `IpAddress` -> `IPAddress`

The formatting looks off. Did you run `go fmt` on your code?

timoreimann:
@tommyvicananza I didn't have a chance to test this yet myself. Did you possibly and can assure that it's working properly?

Another question: Are the new fields read-only?

timoreimann:
Thanks for the contribution! Left some nit-picking notes. :-)

application.go
tommyvicananza: You're totally right about syntax (I did it at work and I don't have a proper installation in this laptop) but I'd to change struct name because it conflicts with another one defined in task.go, what do you think is a proper name, e.g. IPAddressApp???

timoreimann: Good point about the name clash.

I'm still trying to make sense of out of Marathon's documentation. Do I understand correctly that the struct must be set/populated in order for the IP-per-task feature to be activated? If so, then I'd probably go with `IPAddressPerTask` as the name.

tommyvicananza:
do you agree new changes in last commit?

application.go
```diff
@@ -34,6 +34,25 @@ type Applications struct {
 	Apps []Application `json:"apps"`
 }
 
+// IPAddressPerTask is used by IP-per-task functionality Funcionality https://mesosphere.github.io/marathon/docs/ip-per-task.html
```
timoreimann: extra _Funcionality_

application.go
```diff
@@ -69,6 +88,7 @@ type Application struct {
 	AcceptedResourceRoles []string            `json:"acceptedResourceRoles,omitempty"`
 	LastTaskFailure       *LastTaskFailure    `json:"lastTaskFailure,omitempty"`
 	Fetch                 []Fetch             `json:"fetch"`
+	IPAddressPerTask			*IPAddressPerTask		`json:"ipAddress,omitempty"`
```
timoreimann: `go fmt` missing?

timoreimann:
I like the rename. Left two more comments.

I also posed two other questions before:

- Did you possibly test the new functionality? Can you successfully (un-)marshal data from/to the new structs?
- Are the fields read-only or read-write? I think the latter but I'm not 100% sure. If they are, then we also want to have some convenience wrappers to make working with pointers more easy.

tommyvicananza:
Sorry I fixed syntax error you mentioned.

I tested it, I did actually it to use in a custom Traefik which uses Docker Networks and I need these fields.

I don't undestand what you mean about RW or Read-Only

timoreimann:
Sorry if I was unclear: My question was whether clients of go-marathon would normally define the new structs and then send everything to Marathon, or if it would exclusively be populated by Marathon for consumers to read from only.

timoreimann:
I finally had some time to go over the upstream documentation of the feature. AFAICS, the `ipAddress` object itself and any of its nested fields can (and partially must) be set by the client. Due to the way JSON serialization in Go works and the fact that Marathon only offers PATCH semantics through the PUT verb, I believe we should pointerize everything so that clients have the possibility to clear settings. This already holds for some but not all of the new fields.

Along with that, we should also have a set of convenience setters. Could you please take a look at how we do this with other fields (such as (top-level) labels) and mimic things for the new structs? We usually just provide a pair of `Empty*` and `Set*` methods. Since `IPAddressPerTask` is hierarchical, I believe it lends itself to defining the setters at the appropriate levels, respectively. For instance, the setter for `IPAddressPerTask` should be attached to `Application` while those for the field's sub-fields (i.e., `Groups`, `Labels`, and `Discovery`) should be hosted at `IPAddressPerTask`.

timoreimann:
Implemented in #223. Thanks for the up front work. :+1:

PR #133, opened 2016-03-16
Support Marathon 0.15.0+ JSON definition for 422 error responses.

timoreimann:
The `errors` poperty key for 422 HTTP error codes was renamed to `details` in Marathon 0.15.0. (See the [diff on Github.com](https://github.com/mesosphere/marathon/compare/v0.14.2...v0.15.0#diff-96942d2d65f1b0af92336d1786dd6e64L260).)  In order to support both old and newer Marathon versions, we duplicate the
details JSON sub-structure in the overall JSON structure under both the old and new JSON property key.

As a minor cleanup along the way, we remove spaces from a space/tab-mixed intentation and align properly in related tests.

PR #165, opened 2016-06-08
Support Marathon 1.0.0-RC1+ JSON definition for 422 error responses

### LINKED ISSUE #164: HTTP error code 422 (Unprocessable Entity) not deserialized correctly
At least for Marathon 1.1.1, go-marathon does not deserialize the error message from a 422 response properly: As found out in #162, the JSON body

```json
{
  "message": "Object is not valid",
  "details": [
    {
      "path": "\/value",
      "errors": [
        "Requested service port 10000 conflicts with a service port in app \/myapp-green2",
        "Requested service port 10000 conflicts with a service port in app \/myapp"
      ]
    }
  ]
}
```

produces the [APIError](https://github.com/gambol99/go-marathon/blob/697a616b3996092969b86950b1e555e3b4391945/error.go#L48-L52)

```
Marathon API error: Object is not valid (attribute '': )
```

We need to update the parsing logic.

timoreimann:
The error structure for 422 HTTP error codes has [changed again in version 1.0.0-rc1](https://github.com/mesosphere/marathon/pull/3387), adapting the same format as the 400 HTTP error response.

This is an "unwanted successor" to #133.

Fixes #164.

### DIFF
--- a/error.go
+++ b/error.go
@@ -109,20 +109,24 @@ func (def *simpleErrDef) errCode() int {
 	return def.code
 }
 
+type detailDescription struct {
+	Path   string   `json:"path"`
+	Errors []string `json:"errors"`
+}
+
+func (d detailDescription) String() string {
+	return fmt.Sprintf("path: '%s' errors: %s", d.Path, strings.Join(d.Errors, ", "))
+}
+
 type badRequestDef struct {
-	Message string `json:"message"`
-	Details []struct {
-		Path   string   `json:"path"`
-		Errors []string `json:"errors"`
-	} `json:"details"`
+	Message string              `json:"message"`
+	Details []detailDescription `json:"details"`
 }
 
 func (def *badRequestDef) message() string {
 	var details []string
 	for _, detail := range def.Details {
-		errDesc := fmt.Sprintf("path: '%s' errors: %s", detail.Path,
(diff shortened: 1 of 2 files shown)


timoreimann:
@gambol99 could you please take a look?

Apologies for the somewhat hard-to-read code but the upstream changes are really subtle and require a bit of trickery to distinguish cleanly. Suggestions for alternatives are welcome of course.

We should also consider removing support for older formats at some point.

error.go
```diff
@@ -109,20 +109,25 @@ func (def *simpleErrDef) errCode() int {
...
 }
 
+type detailDescription struct {
+	Path   string   `json:"path"`
+	Errors []string `json:"errors"`
+}
+
+func (d detailDescription) String() string {
+	return fmt.Sprintf("path: '%s' errors: %s", d.Path,
+		strings.Join(d.Errors, ", "))
```
gambol99: nit-pick ... do we need the newline?

gambol99:
other than the one picky comment :-) .. looks good to me ..

error.go
timoreimann: Definitely not! Removed.

timoreimann:
Thanks 👍 

Waiting for CI to go green, will merge afterwards.

PR #168, opened 2016-06-15
For Issue-167, add opts to functions like Application.

xinxian0458:
I will complete test, example and document if you think the code is ok.

### DIFF
--- a/application.go
+++ b/application.go
@@ -498,6 +498,24 @@ func (r *marathonClient) Application(name string, opts *GetAppOpts) (*Applicatio
 	return wrapper.Application, nil
 }
 
+// Application retrieves the application configuration from marathon
+// 		name: 		the id used to identify the application
+func (r *marathonClient) ApplicationBy(name string, opts *GetAppOpts) (*Application, error) {
+	u, err := addOptions(buildURI(name), opts)
+	if err != nil {
+		return nil, err
+	}
+	var wrapper struct {
+		Application *Application `json:"app"`
+	}
+
+	if err := r.apiGet(u, nil, &wrapper); err != nil {
+		return nil, err
+	}
+
+	return wrapper.Application, nil
+}
+
 // ApplicationByVersion retrieves the application configuration from marathon
 // 		name: 		the id used to identify the application
 // 		version:  the version of the configuration you would like to receive
@@ -603,14 +621,11 @@ func (r *marathonClient) appExistAndRunning(name string) bool {
 
 // DeleteApplication deletes an application from marathon
 // 		name: 		the id used to identify the application
-func (r *marathonClient) DeleteApplication(name string, opts *DeleteAppOpts) (*DeploymentID, error) {
-	u, err := addOptions(buildURI(name), opts)
-	if err != nil {
-		return nil, err
-	}
+func (r *marathonClient) DeleteApplication(name string, force bool) (*DeploymentID, error) {
+	uri := buildURIWithForceParam(name, force)
 	// step: check of the application already exists
 	deployID := new(DeploymentID)
-	if err := r.apiDelete(u, nil, deployID); err != nil {
+	if err := r.apiDelete(uri, nil, deployID); err != nil {
 		return nil, err
 	}
 
--- a/application.go
+++ b/application.go
@@ -102,7 +102,7 @@ type Fetch struct {
 //		embed:	Embeds nested resources that match the supplied path.
 // 				You can specify this parameter multiple times with different values
 type GetAppOpts struct {
-	Id    string   `url:"id,omitempty"`
+	ID    string   `url:"id,omitempty"`
 	Cmd   string   `url:"cmd,omitempty"`
 	Label string   `url:"label,omitempty"`
 	Embed []string `url:"embed,omitempty"`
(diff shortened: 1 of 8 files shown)


timoreimann:
@xinxian0458, sorry for the late response. A few questions/remarks of mine:

1. I'm uncertain whether we should extend the existing methods by the new filter options struct or instead create a set of new methods (e.g., `ApplicationsBy`, `FilteredApplications`, or similar). My personal thinking is that filtering is rather uncommon and that we should have a separate set of methods, but I'm biased by personal usage experiences. Clearly though, not having to pass an extra `nil` value will be beneficial. Any thoughts, @gambol99?
1. I think we should provide a set of convenience methods for the `embed` option so that clients don't have to build and concatenate string values themselves. Same holds for the various label-specific operators (equal, in, not in, etc.).
1. Doesn't the new `id` field reference the URL parameter we are already setting on operations targeting a specific app only (e.g., `Application(string)`)? Or does it serve a separate purpose I fail to recognize?

gambol99:
> I'm uncertain whether we should extend the existing methods by the new filter options struct or 
> instead create a set of new methods (e.g., ApplicationsBy, FilteredApplications, or similar). My
> personal thinking is that filtering is rather uncommon and that we should have a separate set of 
> methods, but I'm biased by personal usage experiences. Clearly though, not having to pass an 
> extra nil value will be beneficial. Any thoughts, @gambol99?

Admittedly, I'd kinda prefer the using named methods rather packing the logic into a single method on this one.

xinxian0458:
@gambol99 , that's ok.

gambol99:
LGTM ...

timoreimann:
You still need to update a number of GoDoc strings. In particular, there are a few `*By` methods that should be equally named in the comment and mention the second parameter.

application.go
```diff
@@ -95,6 +95,28 @@ type Fetch struct {
...
 }
 
+// GetAppOpts contains a payload for Application method
+//		id:		Filter the result to only return apps whose id is or contains the given value
+//		cmd:	Filter the result to only return apps whose cmd field contains the given value
+//		label:	A label selector query contains one or more label selectors, which are comma separated.
+//		embed:	Embeds nested resources that match the supplied path.
+// 				You can specify this parameter multiple times with different values
+type GetAppOpts struct {
+	Id    string   `url:"id,omitempty"`
```
timoreimann: What's the purpose of this failed? Is it different from the URL parameter that Marathon often requires?

Also, being an [initialism](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) it should be named `ID` .

group.go
```diff
@@ -37,6 +37,31 @@ type Groups struct {
...
+//		embed:		Embeds nested resources that match the supplied path.
+// 					You can specify this parameter multiple times with different values
+type GetGroupOpts struct {
+	Embed []string `url:"embed,omitempty"`
+}
+
+// DeleteGroupOpts contains a payload for DeleteGroup method
+//		force:		Only one deployment can be applied to one application at the same time.
+//					If the existing deployment should be canceled by this change, you can set force=true
+//					Caution: setting force=true xill cancel the current deployment. This parameter should be used only,
```
timoreimann: I'd opt for dropping the implication of what a force means. It's actually a bit [more complicated](https://github.com/mesosphere/marathon/issues/1346) than what your description says. (FWIW, in my organization we force deployments by default since we're not using groups).

My suggestion would be to just say (stolen from the REST docs):

`force:    overrides a currently running deployment.`

timoreimann:
Left a few more comments and a question. Once those are sorted out, I'm fine with things. :-)

application.go
xinxian0458: I used to want to use `GetAppOpts `both in get one application function and in get all applications function, the field of `Id, Cmd, Label` are the query parameter of marathon's `GET /app` api.

group.go
xinxian0458: OK, I can change the comments.

application.go
timoreimann: Okay, makes sense. Just rename it to `ID` then please.

application.go
```diff
@@ -95,6 +95,28 @@ type Fetch struct {
...
+	Cmd   string   `url:"cmd,omitempty"`
+	Label string   `url:"label,omitempty"`
+	Embed []string `url:"embed,omitempty"`
+}
+
+// DeleteAppOpts contains a payload for DeleteApplication method
+//		force:		Only one deployment can be applied to one application at the same time.
+//					If the existing deployment should be canceled by this change, you can set force=true
+//					Caution: setting force=true xill cancel the current deployment. This parameter should be used only,
+//							if the current deployment is unsuccessful!
```
timoreimann: This one must still be reduced.

group.go
```diff
@@ -84,6 +103,35 @@ func (r *marathonClient) Group(name string) (*Group, error) {
 	return group, nil
 }
 
+// Groups retrieves a list of all the groups from marathon by embed options
```
timoreimann: Should say `GroupsBy`.

application.go
```diff
@@ -472,6 +494,25 @@ func (r *marathonClient) Application(name string) (*Application, error) {
 	return wrapper.Application, nil
 }
 
+// Application retrieves the application configuration from marathon
```
timoreimann: Should say `ApplicationBy`.

group.go
```diff
@@ -84,6 +103,35 @@ func (r *marathonClient) Group(name string) (*Group, error) {
...
+		return nil, err
+	}
+	groups := new(Groups)
+	if err := r.apiGet(u, "", groups); err != nil {
+		return nil, err
+	}
+	return groups, nil
+}
+
+// Group retrieves the configuration of a specific group from marathon
```
timoreimann: Should say `GroupBy`.

timoreimann:
LGTM!

timoreimann:
Appreciated @xinxian0458 👍

PR #169, opened 2016-06-29
Add omitempty tag modifiers to Application.Uris and Application.Fetch.

timoreimann:
This fixes the following problems:

- Missing URIs do not cause the `uris` field value to become `null` anymore; while that value currently has the intended effect, we shouldn't rely on it in the long term.
- Same for Fetch URIs (`fetch`).
- Fetch URIs can now be emptied through a new pair of convenience methods.

(Canonical reference: https://willnorris.com/2014/05/go-rest-apis-and-pointers)

Prior, related art: #131


timoreimann:
@gambol99, PTAL.

timoreimann:
Some late local testing of mine seems to indicate that the given PR doesn't solve the problem I thought I had.

I'm putting this PR on hold to double check. @gambol99, nothing to see here yet. :-)

timoreimann:
Nevermind, I let myself get side-tracked for no reason. I think we need this PR. :-)

@gambol99, it's all yours!

gambol99:
couldn't find fault :-) ... LGTM

PR #174, opened 2016-07-21
Mark app as unhealthy if any health check result is missing.

### LINKED ISSUE #118: By design? ApplicationOk returns true for health checks without results
Is it by design that `ApplicationOk` returns `true` if an application has a health check defined but the health check result being `nil`?

We're talking about this section of code here:
```go
	// step: iterate the application checks and look for false
	for _, task := range application.Tasks {
		if task.HealthCheckResult != nil {
			for _, check := range task.HealthCheckResult {
				//When a task is flapping in Marathon, this is sometimes nil
				if check == nil {
					return false, nil
				}
				if !check.Alive {
					return false, nil
				}
			}
		}
	}
```

What I am trying to accomplish is to implement a zero-downtime deployment where an older version of an app (e.g. `/my/app/v1`) should continue to run until the new one (e.g. `/my/app/v2`) has been deployed successfully with all health checks passing.
Polling `ApplicationOk` until that returns `true` does not work, because of the behaviour stated above: For the longest time during a deployment of a new application version, all `HealthCheckResults` of its `Tasks` are `nil` - and therefore `ApplicationOk` returns `true` before the new app version is initialised or even up.

My question is: Am I holding it wrong? ;)
I mean is there a different and better way you designed to do this?
Or is the behaviour described above simply a corner case not accounted for just yet?

timoreimann:
During deployments, Marathon does not immediately return health check results. We need to play it safe here and consider such apps as unhealthy.

Fixes #118.

### DIFF
--- a/application.go
+++ b/application.go
@@ -563,6 +563,12 @@ func (r *marathonClient) ApplicationOK(name string) (bool, error) {
 
 	// step: iterate the application checks and look for false
 	for _, task := range application.Tasks {
+		// Health check results may not be available immediately. Assume
+		// non-healthiness if they are missing for any task.
+		if task.HealthCheckResults == nil {
+			return false, nil
+		}
+
 		for _, check := range task.HealthCheckResults {
 			//When a task is flapping in Marathon, this is sometimes nil
 			if check == nil || !check.Alive {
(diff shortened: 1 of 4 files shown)


timoreimann:
@gambol99 PTAL.

/cc @oolongbrothers @iderdik

iderdik:
LGTM :)

timoreimann:
Thanks -- going to merge this now.

PR #176, opened 2016-07-27
Fix bug: can not unmarshal EventDeploymentInfo.CurrentStep

### LINKED ISSUE #175: Can not unmarshal the attribute [CurrentStep] in Event [Deployment Info] to object EventDeploymentInfo
either can EventDeploymentStepSuccess and EventDeploymentStepFailure

Taking EventDeploymentInfo for example, the event response message is:
``` json
....
	"currentStep": {
		"actions": [
			{
				"type": "ScaleApplication",
				"app": "/testorg/testspace/test.app"
			}
		]
	},
....
```
But the definition of EventDeploymentInfo is:
``` go
// EventDeploymentInfo describes a 'deployment_info' event.
type EventDeploymentInfo struct {
	EventType   string          `json:"eventType"`
	CurrentStep *DeploymentStep `json:"currentStep"`
	Timestamp   string          `json:"timestamp"`
	Plan        *DeploymentPlan `json:"plan"`
}

// DeploymentStep is a step in the application deployment plan
type DeploymentStep struct {
	Action string `json:"action"`
	App    string `json:"app"`
}
```
They do not match.

wu8685:
fix #175


timoreimann:
Thanks for the PR. I still need to find some time to go through this, will probably take a few more days.

What we should most likely have though is a regression test. @wu8685, could you consider adding one please?

wu8685:
ok, I think it's necessary.

deployment.go
```diff
@@ -56,23 +56,11 @@ type DeploymentStepV1 struct {
...
-		Dependencies []string       `json:"dependencies"`
-		Groups       []*Group       `json:"groups"`
-		ID           string         `json:"id"`
-		Version      string         `json:"version"`
-	} `json:"target"`
+	ID       string            `json:"id"`
+	Version  string            `json:"version"`
+	Original *Group            `json:"original"`
+	Target   *Group            `json:"target"`
+	Steps    []*DeploymentStep `json:"steps"`
```
timoreimann: This change isn't strictly related to the fix, is it?

If not, then I'd prefer to have it as a separate PR in order to keep this PR focused. I'm happy to let you do it or take care of it myself.

events.go
```diff
@@ -286,26 +286,26 @@ type EventDeploymentFailed struct {
 
 // EventDeploymentInfo describes a 'deployment_info' event.
 type EventDeploymentInfo struct {
-	EventType   string          `json:"eventType"`
-	CurrentStep *DeploymentStep `json:"currentStep"`
-	Timestamp   string          `json:"timestamp"`
-	Plan        *DeploymentPlan `json:"plan"`
+	EventType   string            `json:"eventType"`
+	CurrentStep *DeploymentStepV1 `json:"currentStep"`
```
timoreimann: The name `DeploymentStepV1` turned out to be misleading as the structure is already used in pre-1.x. (See also [my comment](https://github.com/gambol99/go-marathon/issues/175#issuecomment-235815659).) We should either rename `DeploymentStepV1` to something more generic that just describes the structure, or rename it to `DeploymentStep` and also rename the _current_ `DeploymentStep`to something different (like `DeploymentStepOld`).

I'll need to ponder about a good choice and good names. Feel free to come up with suggestions too. :-)

deployment.go
wu8685: Yes, it's just a refactoring having nothing to do with this issue.
If it is unnecessary, I will remove it.

events.go
wu8685: I think we should keep `DeploymentStepV1` as it is, because it is also used to parse [deployments](https://github.com/gambol99/go-marathon/blob/master/deployment.go#L91) for compatibility. We could clone `DeploymentStepV1` to a new struct.

As for the name, how about `DeploymentCurrentStep` ? I think `CurrentStep` is also ok, as this struct is just used in `EventDeploymentInfo`, `EventDeploymentStepSuccess` and `EventDeploymentStepFailure` under events.go and we don't need to tell it's a deployment step.

deployment.go
wu8685: Shall we reconsider this refactoring : ) ? I encounter a scenario like the following to reference this "Group":
``` go
findAppInGroup(e.Plan.Original, targetAppId)

func findAppInGroup(group *gomarathon.Group, id string) {
	......
}
```

In current code, I will have to declare the function with the `group` being `interface{}`. It's not convenient for the following use.

timoreimann: I do want the refactoring to happen but would rather have it as a separate PR. Just pushed #177 for that purpose.

Not 100% I understand your concern about `group` being `interface{}`. Do I see correctly that merging #177 will solve that problem?

events.go
timoreimann: Renaming `DeploymentStepV1`to `CurrentStep` sounds like a good choice to me.

Just to be sure we're on the same page: I understand we'll use that struct in the event types you have mentioned and the deployments unmarshaling for Marathon version 1.x (i.e., [here](https://github.com/gambol99/go-marathon/blob/2e73c1d824745b3b6374f2d7cc46932f5a596698/deployment.go#L91)).

deployment.go
timoreimann: I merged the PR.

events.go
wu8685: I'm not sure if we should just use `CurrentStep` to replace `DeploymentStepV1` in [unmarshaling deployments](https://github.com/gambol99/go-marathon/blob/2e73c1d824745b3b6374f2d7cc46932f5a596698/deployment.go#L91). I believe a `CurrentStep` placed there would be confusing. Following is my options:

1. Keep the `DeploymentStepV1` as it is and we don't need to change anything more. Because `EventDeploymentInfo.CurrentStep` is surely one kind of deployment step.
2. If option 1 makes some confusion, we can copy `DeploymentStepV1` to a new struct and name it `CurrentStep`. Then we use this new `CurrentStep` in `EventDeploymentInfo.CurrentStep`, and keep `DeploymentStepV1` in [unmarshaling deployments](https://github.com/gambol99/go-marathon/blob/2e73c1d824745b3b6374f2d7cc46932f5a596698/deployment.go#L91). I mentioned the reason at the beginning. (This is what I mentioned in my last comment. )
3. Maybe we could make some refactoring in [unmarshaling deployments](https://github.com/gambol99/go-marathon/blob/2e73c1d824745b3b6374f2d7cc46932f5a596698/deployment.go#L89-L105). Considering `DeploymentStepV1` is later than `DeploymentStep`, we could reverse their place. We could try to unmarshal `DeploymentStepV1` at first. If failed, we then unmarshal `deployment.XXStepsRaw` to `DeploymentStep`. After that, we can name current `DeploymentStepV1` as `DeploymentStep` and current `DeploymentStep` as `DeploymentStepV0` or something else, and make `EventDeploymentInfo.CurrentStep` is type `DeploymentStep` which is `DeploymentStepV1` before. I think this would be more reasonable.

wu8685:
I add a test case for deployment info

events.go
timoreimann: I was about to ask you why you'd find a rename to `CurrentStep` confusing when I realized that for the event types, `currentStep` (the JSON property) is the kind of object we've been talking about mostly whereas in the `deployments` endpoint response, the same property denotes an integer. That's an unfortunate choice on Marathon's end in my opinion.

Regarding option 1: As you suspected, I'd rather not keep using the `V1` suffix as we figured out that `DeploymentStepV1` (being used for events already) is not specific to Marathon 1.x at all, so it'd give a wrong impression and likely lead to confusion. We'd probably want to rename it again at some point in the future, which is going to be another (albeit small) breaking change.

Regarding option 3: While I like this approach in general, AFAICS it means that we'd need to change the type that `Deployment.Steps` points to. That's something we deliberately avoided during #158 because we were not sure if the breaking API change described in [upstream issue 3965](github.com/mesosphere/marathon/issues/3965) would be reverted or not. The labelling on that issue seems to indicate that things will be reverted; I [asked for clarification again earlier today](https://github.com/mesosphere/marathon/issues/3965#issuecomment-236159923) and hope to get some feedback. Until then, I'd like to maintain the externally visible structure of the `Deployment` struct and not suffix a type with `V0` when there's still a possibility it will eventually become the canonical type for Marathon version 1.

Regarding option 2: I believe that `EventDeployment.*.CurrentStep` and `Deployment.Steps` are supposed to reference the same type, so duplicating `DeploymentStepV1` under a different name and decoupling the two may lead to additional maintenance effort. My suggestion would be to rename the struct `DeploymentStepV1` to something like `DeploymentStepActions` or `StepActions` (i.e., describe what it actually is instead of trying to match its corresponding field name) and use that for the events and in [marathon.Deployments()](https://github.com/gambol99/go-marathon/blob/2e73c1d824745b3b6374f2d7cc46932f5a596698/deployment.go#L91). Since `DeploymentStepV1` is only used internally, we wouldn't break anyone now nor later if upstream decides to fix the breaking change by means of reverting.

So my opinion would be to go with option 2 but rename instead of clone/duplicate.

What do you think?

wu8685: It's acceptable. I have changed the name to `StepActions`

deployment.go
```diff
@@ -47,7 +47,7 @@ type DeploymentStep struct {
 }
 
 // DeploymentStepV1 is a Marathon v1.X-compatible DeploymentStep
```
timoreimann: The name in the comment needs to be updated. I'd also suggest a more generic descriptions, such as:

_StepActions is a series of deployment steps._

events.go
```diff
@@ -286,26 +286,26 @@ type EventDeploymentFailed struct {
...
 // EventDeploymentInfo describes a 'deployment_info' event.
 type EventDeploymentInfo struct {
-	EventType   string          `json:"eventType"`
-	CurrentStep *DeploymentStep `json:"currentStep"`
-	Timestamp   string          `json:"timestamp"`
-	Plan        *DeploymentPlan `json:"plan"`
+	EventType   string            `json:"eventType"`
+	CurrentStep *StepActions `json:"currentStep"`
+	Timestamp   string            `json:"timestamp"`
+	Plan        *DeploymentPlan   `json:"plan"`
```
timoreimann: It looks like `gofmt` didn't run on here and on the other blocks in `events.go`.

wu8685:
I have corrected them.

timoreimann:
LGTM. Thanks for digging into the problem!

PR #177, opened 2016-07-28
Use Group struct in DeploymentPlan.

timoreimann:
Spawn off of #176, specifically [this comment](https://github.com/gambol99/go-marathon/pull/176#discussion-diff-72573214).

/cc @wu8685


wu8685:
It's ok to me.

Issue #179, opened 2016-08-10
Cluster members may have URL prefix


fsamin:
Hello,

First of all, thanks for your work !

It looks like #178, but i'm not sure it's the same use case.
My Mesos hosts are hosted behind a reverse proxy (`https://<my-host>/marathon/<foo>`). 
`newCluster` function (https://github.com/gambol99/go-marathon/blob/master/cluster.go#L109) ignores path. 

So I fixed it with this : 
 - https://github.com/fsamin/go-marathon/blob/fix-cluster-members-url/cluster.go#L109

Let me know if you want me to fix it a different way.

Regards
François

PR #180, opened 2016-08-10
Fix: cluster members may have url prefix

fsamin:
see #179


gambol99:
cheers for this @fsamin ... could be we have a quick test for with and without paths ... perhaps based off to https://github.com/gambol99/go-marathon/blob/master/cluster_test.go#L47

fsamin:
No problem @gambol99, I'll write this test.

fsamin:
Hi @gambol99, I add a test with a path. URL without Path were already tested.
Regards

gambol99:
lgtm

PR #187, opened 2016-08-23
Add message as a field in EventStatusUpdate

mikesol314:
According to the Marathon API docs (https://mesosphere.github.io/marathon/docs/generated/api.html#v2_events_get), a message field is included in the status_update_event. We need this message specifically for when a task fails as it gives a failure reason. This change updates the EventStatusUpdate structure to include said message field.


timoreimann:
LGTM.

Issue #190, opened 2016-08-31
Docker Container PortMapping Labels


catherinetcai:
Currently, go-marathon doesn't support labels for Docker container portmappings: 

```
"container": {
    "docker": {
      "forcePullImage": false,
      "image": "mesosphere:marathon/latest",
      "network": "BRIDGE",
      "parameters": [
        {
          "key": "name",
          "value": "kdc"
        }
      ],
      "portMappings": [
        {
          "containerPort": 80,
          "hostPort": 0,
          "protocol": "tcp",
          "servicePort": 10019,
          "name": "http",
          **"labels": { "vip": "192.168.0.1:80" }**
        }
      ],
      "privileged": false
    }
```

I think a map should be added on the PortMapping struct to allow for labels to be passed in. These labels are useful for setting VIPs in DCOS.

If you guys find this useful, I already have a fork of go-marathon with these changes.

timoreimann:
Hey @catherinetcai, I believe @Mongey's PR #191 fixed this. I'll close the issue -- if you believe there's still something missing, please don't hesitate to ask for reopening.

Thanks!

PR #191, opened 2016-09-01
Add name & labels to container portMapping

Mongey:


docker.go
```diff
@@ -30,10 +30,12 @@ type Container struct {
 
 // PortMapping is the portmapping structure between container and mesos
 type PortMapping struct {
-	ContainerPort int    `json:"containerPort,omitempty"`
-	HostPort      int    `json:"hostPort"`
-	ServicePort   int    `json:"servicePort,omitempty"`
-	Protocol      string `json:"protocol,omitempty"`
+	ContainerPort int                `json:"containerPort,omitempty"`
+	HostPort      int                `json:"hostPort"`
+	Labels        *map[string]string `json:"labels,omitempty"`
```
timoreimann: We normally provide convenience methods to easily populate pointerized variables. For labels, we already have something to [add](https://github.com/gambol99/go-marathon/blob/63fea16e82279c1d934bcb4602614308b69f22c3/application.go#L238) and [clear labels](https://github.com/gambol99/go-marathon/blob/63fea16e82279c1d934bcb4602614308b69f22c3/application.go#L250). I think we should duplicate these functions on `PortMapping`.

(I initially thought we could provide a single implementation on a custom `map[string]string` type but came to believe this won't work due to the fact that we can't initialize a receiver map on a receiver method.)

timoreimann:
@Mongey thanks for the PR. Left one remark.

timoreimann:
I noticed something else: We should also update [ExposePort](https://github.com/Mongey/go-marathon/blob/6f861d3325a02d32170f5c2c393acaeab72c683e/docker.go#L162) so that the new port mapping values can be easily attached to a `Docker` instance. My suggestion would be to replace the four individual parameters by a single `PortMapping` instance. As an additional benefit, callers will be able to pass in just the `PortMapping` fields they are interested in.

`Expose` and `ExposeUDP` will have to be updated along the way. This should be trivial, though.

timoreimann:
Could you please remove (i.e., git amend and force-push) the `Fixes #190` part from commit 8fcf971 to prevent multiple references on the related issue, especially as we continue to work on this PR?

It's usually better to reference issues via the pull request description.

Mongey:
@timoreimann thanks for the feedback. Pushed up some changes

timoreimann:
@Mongey Would you mind adding a quick test for the new `PortMapping` label convenience methods? I think we can pretty much duplicate [TestApplicationLabels](https://github.com/gambol99/go-marathon/blob/0d84001b1247ffd4295c644b986db075395d1a3e/application_test.go#L137). (Ideally, I'd like one test to cover both label types; not sure though how to achieve that elegantly given that the convenience methods are defined on different types.)

After that, I think we can merge your PR. 👍

timoreimann:
Great, LGTM.

Thanks again, @Mongey. 👏

Mongey:
🙇

PR #193, opened 2016-09-05
Fix timeout error in WaitOnApplication

ch3lo:
timer.After is a function who returns a new channel, this causes a problem because the timer is resetting in each iteration.

You can see a detailed explain in http://stackoverflow.com/questions/35036653/why-doesnt-this-golang-code-to-select-among-multiple-time-after-channels-work/35036775

Regards

### DIFF
--- a/application_test.go
+++ b/application_test.go
@@ -458,19 +458,56 @@ func TestApplicationConfiguration(t *testing.T) {
 	assert.Equal(t, ErrCodeNotFound, apiErr.ErrCode)
 }
 
-func TestWaitOnApplication(t *testing.T) {
+func TestWaitOnApplicationWithTimeoutGreaterThanTicker(t *testing.T) {
+	defaultConfig := NewDefaultConfig()
+	configs := &configContainer{}
+	configs.client = &defaultConfig
+	configs.client.WaitTime = 100
+
+	endpoint := newFakeMarathonEndpoint(t, configs)
+	defer endpoint.Close()
+
+	var err error
+	go func() {
+		err = endpoint.Client.WaitOnApplication(fakeAppName, 200*time.Millisecond)
+	}()
+	timer := time.NewTimer(500*time.Millisecond)
+	<- timer.C
+	assert.NoError(t, err)
+
+
+	err = nil
+	go func() {
+		err = endpoint.Client.WaitOnApplication("no_such_app", 200*time.Millisecond)
+	}()
+	timer = time.NewTimer(500*time.Millisecond)
+	<- timer.C
+	assert.IsType(t, err, ErrTimeoutError)
+}
+
+func TestWaitOnApplicationWithTimeoutSmallerThanTicker(t *testing.T) {
+	defaultConfig := NewDefaultConfig()
+	configs := &configContainer{}
+	configs.client = &defaultConfig
+	configs.client.WaitTime = 200
+
 	endpoint := newFakeMarathonEndpoint(t, nil)
 	defer endpoint.Close()
 
 	var err error
-	err = endpoint.Client.WaitOnApplication(fakeAppName, 1*time.Second)
+	go func() {
+		err = endpoint.Client.WaitOnApplication(fakeAppName, 100*time.Millisecond)
+	}()
+	timer := time.NewTimer(500*time.Millisecond)
+	<- timer.C
 	assert.NoError(t, err)
+
 
 	err = nil
 	go func() {
-		err = endpoint.Client.WaitOnApplication("no_such_app", 1*time.Second)
+		err = endpoint.Client.WaitOnApplication("no_such_app", 100*time.Millisecond)
 	}()
-	timer := time.NewTimer(1200*time.Millisecond)
+	timer = time.NewTimer(500*time.Millisecond)
 	<- timer.C
 	assert.IsType(t, err, ErrTimeoutError)
 }
--- a/application_test.go
+++ b/application_test.go
@@ -466,9 +466,15 @@ func TestWaitOnApplication(t *testing.T) {
 	err = endpoint.Client.WaitOnApplication(fakeAppName, 1*time.Second)
 	assert.NoError(t, err)
 
-	err = endpoint.Client.WaitOnApplication("no_such_app", 1*time.Millisecond)
+	err = nil
+	go func() {
+		err = endpoint.Client.WaitOnApplication("no_such_app", 1*time.Second)
+	}()
+	timer := time.NewTimer(1200*time.Millisecond)
+	<- timer.C
 	assert.IsType(t, err, ErrTimeoutError)
 }
+
 func TestAppExistAndRunning(t *testing.T) {
 	endpoint := newFakeMarathonEndpoint(t, nil)
 	defer endpoint.Close()
--- a/application_test.go
+++ b/application_test.go
@@ -458,58 +458,58 @@ func TestApplicationConfiguration(t *testing.T) {
 	assert.Equal(t, ErrCodeNotFound, apiErr.ErrCode)
 }
 
-func TestWaitOnApplicationWithTimeoutGreaterThanTicker(t *testing.T) {
-	defaultConfig := NewDefaultConfig()
-	configs := &configContainer{}
-	configs.client = &defaultConfig
-	configs.client.WaitTime = 100
+func TestWaitOnApplication(t *testing.T) {
+	waitTime := 100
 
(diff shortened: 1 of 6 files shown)


gambol99:
.... strange this one wasn't been caught already! ...

LGTM .. @timoreimann?

timoreimann:
@gambol99 maybe `WaitOnApplication` isn't so widely used. Or, it's just that the Marathon deployments out there are mostly stable and tend to not run into the timeout. :-)

(Coincidentally, I ran into a variation of this problem at work just a week ago.)

@ch3lo, thanks for your PR -- LGTM!

timoreimann:
Ah, maybe one thing: Does this hint at a problem with our testing?

timoreimann:
I think [our test](https://github.com/ch3lo/go-marathon/blob/c543ea43702776768edcb38f7a2914f569c7063d/application_test.go#L461-L462) is flawed in the sense that the timeout is chosen smaller than the ticker period, so we end up never iterating through the for loop more than once. Passing in a timeout higher than the ticker period shows that the test (in master) runs infinitely.

My suggestion would be to extract the ticker value into a constant and update the two test cases such that we pick a timeout lower and higher than the ticker value for the happy and unhappy path, respectively (e.g., double and half the timeout).

@gambol99, WDYT?

ch3lo:
Hi all, thanks for your prompt response.

The unit test did not fail because we need a timeout greater than the ticker (500ms). So, I wrote a unit test that fails if the deploy takes more time than the specified in the timeout parameter.

What do you think?

timoreimann:
Ah yeah, wrapping the method under test in a goroutine is probably a good idea. Could you do the same with the happy path test for consistency and safety reasons?

It's a bit hard to understand why we choose the respective timeouts. What do you think about extracting  the ticker timeout into a constant and referencing it in the tests as suggested in my previous comment?

application_test.go
```diff
@@ -450,17 +450,60 @@ func TestApplicationConfiguration(t *testing.T) {
 	assert.Equal(t, ErrCodeNotFound, apiErr.ErrCode)
 }
 
-func TestWaitOnApplication(t *testing.T) {
+func TestWaitOnApplicationWithTimeoutGreaterThanTicker(t *testing.T) {
+	defaultConfig := NewDefaultConfig()
+	configs := &configContainer{}
+	configs.client = &defaultConfig
+	configs.client.WaitTime = 100
```
timoreimann: I find it more readable to set `WaitTime` right after `defaultConfig` has been defined. Moreover, I would use literal structs to be slightly less repetitive; so overall:

```
defaultConfig := NewDefaultConfig()
defaultConfig.WaitTime = 100
configs := &configContainer{
  client: &defaultConfig,
}
```

application_test.go
```diff
@@ -450,17 +450,60 @@ func TestApplicationConfiguration(t *testing.T) {
...
+	go func() {
+		err = endpoint.Client.WaitOnApplication(fakeAppName, 100*time.Millisecond)
+	}()
+	timer := time.NewTimer(500*time.Millisecond)
+	<- timer.C
 	assert.NoError(t, err)
 
-	err = endpoint.Client.WaitOnApplication("no_such_app", 1*time.Millisecond)
+
+	err = nil
```
timoreimann: I think we should play safe here and use a new error variable to rule out interferences between the previous invocation of `WaitOnApplication` (line 491) and the following one (line 500) in case of a broken implementation.

However, see also my comment on converting both new test functions into a single, table-driven test.

application_test.go
```diff
@@ -450,17 +450,60 @@ func TestApplicationConfiguration(t *testing.T) {
...
+	go func() {
+		err = endpoint.Client.WaitOnApplication("no_such_app", 100*time.Millisecond)
+	}()
+	timer = time.NewTimer(500*time.Millisecond)
+	<- timer.C
 	assert.IsType(t, err, ErrTimeoutError)
 }
```
timoreimann: `TestWaitOnApplicationWithTimeoutGreaterThanTicker` and `TestWaitOnApplicationWithTimeoutSmallerThanTicker` seem exactly identical except for the timeout values chosen. I believe we could combine the two into a single, table-driven test.

Here's what I came up with:

```
func TestWaitOnApplication(t *testing.T) {
	waitTime := 100 * time.Millisecond

	tests := []struct {
		desc    string
		timeout time.Duration
		appName string
	}{
		{
			desc:    "existing app / timeout > ticker",
			timeout: 200 * time.Millisecond,
			appName: fakeAppName,
		},
		{
			desc:    "missing app / timeout > ticker",
			timeout: 200 * time.Millisecond,
			appName: "no_such_app",
		},
		{
			desc:    "existing app / timeout < ticker",
			timeout: 50 * time.Millisecond,
			appName: fakeAppName,
		},
		{
			desc:    "missing app / timeout < ticker",
			timeout: 50 * time.Millisecond,
			appName: "no_such_app",
		},
	}

	for _, test := range tests {
		defaultConfig := NewDefaultConfig()
		defaultConfig.WaitTime = waitTime
		configs := &configContainer{
			client: &defaultConfig,
		}

		endpoint := newFakeMarathonEndpoint(t, configs)
		defer endpoint.Close()

		var err error
		go func() {
			err = endpoint.Client.WaitOnApplication(test.appName, test.timeout)
		}()
		timer := time.NewTimer(400 * time.Millisecond)
		<-timer.C
		if test.appName == fakeAppName {
			assert.NoError(t, err, test.desc)
		} else {
			assert.IsType(t, err, ErrTimeoutError, test.desc)
		}
	}
}
```

This also elegantly resolves my concern about goroutine interferences by means of scoping the `error` variable to its test case iteration.

client.go
```diff
@@ -175,6 +175,8 @@ type marathonClient struct {
 	listeners map[EventsChannel]int
 	// a custom logger for debug log messages
 	debugLog *log.Logger
+	// wait time between repetitive requests to API
+	waitTime time.Duration
```
timoreimann: Maybe be a bit more specific and call it `pollingWaitTime`. Would then also update the description to

`//  wait time between repetitive requests to the API during polling`

Likewise in the `Config` definition.

config.go
```diff
@@ -47,6 +47,8 @@ type Config struct {
 	LogOutput io.Writer
 	// HTTPClient is the http client
 	HTTPClient *http.Client
+	// WaitTime time in milliseconds
```
timoreimann: Let's copy the description from the `marathonClient` struct here.

config.go
```diff
@@ -47,6 +47,8 @@ type Config struct {
...
+	WaitTime int
```
timoreimann: Does anything speak against making this of type `time.Duration` as well?

timoreimann:
I think we should also be able to use the configured wait time here: https://github.com/ch3lo/go-marathon/blob/38f690c0e9945069b194e37e4ef41c8068ee1ef8/group.go#L198

timoreimann:
Could you please make sure you run gofmt over your PR? At least the new code in `application_test.go` does not seem to be formatted correctly.

config.go
```diff
@@ -47,6 +47,8 @@ type Config struct {
 	LogOutput io.Writer
 	// HTTPClient is the http client
 	HTTPClient *http.Client
+	// wait time (in milliseconds) between repetitive requests to the API during polling
+	PollingWaitTime int
```
timoreimann: Would type this as `time.Duration` to maximize flexibility for users and be consistent with `marathonClient`.

config.go
```diff
@@ -57,5 +59,6 @@ func NewDefaultConfig() Config {
 		EventsPort:      10001,
 		EventsInterface: "eth0",
 		LogOutput:       ioutil.Discard,
+		PollingWaitTime: 500,
```
timoreimann: If we retype `PollingWaitTime`, we should do `500 * time.Millisecond` here. (I know that milliseconds is default, but typing it out (pun intended) makes things more clear.)

timoreimann:
Hey @ch3lo, looking pretty good now. Have one more tinyish usability concern, then I'm happy. :-)

ch3lo:
Nice, thank you for your suggestions. Now I'm writing table driven tests in my internal projects and they look more clear.

Cheers

config.go
timoreimann: @ch3lo, do you think we could address this point? It's the last piece missing before we can merge.

Thanks!

timoreimann:
Carried the PR along in PR 202 due to inactivity.

Thanks for your contribution -- appreciated!

PR #196, opened 2016-09-13
Add port definitions to application.

taotetek:
Application details currently provides the list of ports for an application, but does not provide port definitions. We are using the "name" field within definitions to denote port functionalities for things such as prometheus (metrics) exporters to aid in service discovery.  This allows us to do things like configure prometheus scrapers to find all metrics exporter ports regardless of which application they are for automatically.

This PR adds the PortDefinitions field the Application struct and adds a PortDefinition type.

go-marathon is great by the way - saving us so much time!


port_definition.go
```diff
@@ -0,0 +1,23 @@
+/*
+
```
timoreimann: I noticed that (most) other files open up the copyright preamble with a `Copyright <year> Rohith All rights reserved` line. My knowledge about software licensing is at a bare minimum, so I'm just guessing that this might be relevant or totally not.

Pinging @gambol99 since he's the one mentioned on the line in question. :-)

port_definition.go
```diff
@@ -0,0 +1,23 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package marathon
+
+// PortDefinition provides metadata about a port
+type PortDefinition struct {
+	Port     int    `json:"port,omitempty"`
```
timoreimann: My understanding of the [upstream documentation](https://mesosphere.github.io/marathon/docs/generated/api.html#v2_apps_post) is that the `port` JSON field carries special semantics when set to zero:

> For every port definition with port number zero, a globally unique (cluster-wide) service port is assigned and provided as part of the app definition to be used in load balancing definitions.

This sounds like we must permit to marshal a 0 (zero) value by means of dropping the `omitempty` tag specifier and turn `Port` into a pointer-to-int so that we can represent the three available states missing, zero, and non-zero. We normally implement convenience methods to clear (nilify) and set pointerized variables.

WDYT?

timoreimann:
@taotetek thanks for the kind words and your PR!

I left two comments.

taotetek:
@timoreimann thanks for the feedback! I did indeed just copy over the boilerplate copyright without reading it too carefully.  I'll address your comments either this evening or first thing in the morning.  Cheers!

port_definition.go
taotetek: @timoreimann nice catch on that - you are correct about port 0 being a special value, will push a fix for this line to the branch.

taotetek:
@timoreimann There we go - I went ahead and modified the copyright preamble to be in line with the other files while fixing the port definition.  Cheers!

port_definition.go
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package marathon
+
+// PortDefinition provides metadata about a port
+type PortDefinition struct {
+	Port     int    `json:"port"`
```
timoreimann: I think we need to turn this into a pointer in order to distinguish between omitting, zero, and non-zero as the three carry different semantics (to my understanding) and Marathon only does PATCH-like updates even though everything is PUT. (For an extended rationale, see [this blog post on Go and REST APIs](https://willnorris.com/2014/05/go-rest-apis-and-pointers).)

Would you mind making the change and adding a convenience setter method on `PortDefinition` similar to [Application.Mem()](https://github.com/gambol99/go-marathon/blob/master/application.go#L175) (plus a quick test)?

application.go
```diff
@@ -50,6 +50,7 @@ type Application struct {
 	Mem                   *float64            `json:"mem,omitempty"`
 	Tasks                 []*Task             `json:"tasks,omitempty"`
 	Ports                 []int               `json:"ports"`
+	PortDefinitions       []*PortDefinition   `json:"portDefinitions,omitempty"`
```
timoreimann: I think this needs to be a `*[]PortDefinition` if we want to be able to submit an empty array (in order to, say, clear an existing array of non-empty port definitions). If you agree, then we should also have a (tested) `AddPortDefinition()` helper method (similar to [AddHealthCheck()](https://github.com/gambol99/go-marathon/blob/5e63f97a03e7c519e489834142de7893649aae53/application.go#L286)) to ease usage of pointers.

(See my other comment regarding why we need pointers in the first place.)

port_definition.go
```diff
@@ -0,0 +1,24 @@
+/*
+Copyright 2014 Rohith All rights reserved.
```
timoreimann: nit-pick: 2016. :-)

timoreimann:
Hey @taotetek, that was quick. :-)

I spotted one or two more things on the delicate matter of when to use pointers in Go REST APIs. Let's work those out and get the PR flying afterwards.

taotetek:
@timoreimann sure thing! will try to get to it this evening, and thanks for the once over! will test changes in our environment after I make them as well.

timoreimann:
Appreciate the extra testing 🎉

taotetek:
@timoreimann just a heads up that I got bogged down with a couple of other things I need to finish - but I will circle back to this once I could a couple of other things off my plate this week. cheers!

taotetek:
@timoreimann Should have this wrapped up tomorrow :) Have everything done on local branch except for support for labels in port definitions which I hope to make time for in the morning.

timoreimann:
@taotetek great -- thanks for the heads up!

taotetek:
@timoreimann latest updates! note I won't have time to test this against a real mesos setup until tomorrow.  cheers!

timoreimann:
Oh by the way, if you feel like you may also add yourself to the AUTHORS file. No obligations though. :-)

application.go
```diff
@@ -179,6 +179,29 @@ func (r *Application) Memory(memory float64) *Application {
 	return r
 }
 
+// AddPortDefinition adds a port definition. Port definitions are used to define ports that
+// should be considered part of a resource. They are necessary when you are using HOST
+// networking and no port mappings are specified.
+func (r *Application) AddPortDefinition(portDefinition PortDefinition) *Application {
+	if r.PortDefinitions == nil {
+		r.PortDefinitions = &[]PortDefinition{}
```
taotetek: @timoreimann ah yep! meant to change that after I added EmptyPortDefinitions in fact - will follow up!

taotetek: Ok, pushed!

taotetek:
@timoreimann will try converting one  of our application definitions over from json file + curl to go-marathon tomorrow and let you know!

taotetek:
@timoreimann just a heads up - tested the code against our mesos / marathon setup - working fine!

timoreimann:
@taotetek glad to hear -- let's get this thing into master then!

Once more, thanks for the PR. Highly appreciated!

PR #200, opened 2016-09-22
add support for persistent vols to containers

taotetek:
This PR adds support for persistent volumes. It has been successfully tested against our mesos / marathon infrastructure.  Feedback welcome!


gambol99:
LGTM

timoreimann:
I'd still like to double check if we could possibly drop a pointer here or there. For instance, are zero-length string-typed names valid consistently? That is, do we need to be able to clear them?

Apart from that, things look fine to me!

docker.go
```diff
@@ -46,9 +46,48 @@ type Parameters struct {
...
+	ContainerPath string            `json:"containerPath,omitempty"`
+	HostPath      string            `json:"hostPath,omitempty"`
+	Mode          string            `json:"mode,omitempty"`
+	Persistent    *PersistentVolume `json:"persistent,omitempty"`
+}
+
+// PeristentVolume declares a Volume to be Persistent, and sets
+// the size (in MiBs) and optional constraints for the Volume.
+type PersistentVolume struct {
+	Type        *string     `json:"type,omitempty"`
```
timoreimann: As there are only the three available types `root`, `path`, and `mount`, how about creating a string type declaration `PersistentVolumeType` and define three constants accordingly? This would make it easier for users to understand the valid input domain and also enable IDE/editor auto-completion.

(FWIW, Kubernetes does this a lot, like [here](https://github.com/kubernetes/kubernetes/blob/5f2846eefb48cf3ec1d1130c64a7007e19430200/pkg/api/v1/types.go#L495-L502)).

I also think we don't need a pointer as the empty string is not a valid option.

docker.go
```diff
@@ -46,9 +46,48 @@ type Parameters struct {
...
+	Size        int         `json:"size,omitempty"`
+	Constraints *[][]string `json:"constraints,omitempty"`
+}
```
timoreimann: Do I see correctly that neither `type` nor `constraints` are [documented in the REST API](https://mesosphere.github.io/marathon/docs/generated/api.html#v2_apps_post)? The relevant JSON blob looks like this:

```json
"persistent":{
  "additionalProperties":false,
  "properties":{
     "size":{
        "type":"integer",
        "description":"The size of the persistent volume in MiB",
        "minimum":0
     }
  },
  "type":"object"
}
```

I can only see a definition in the [documentation on local persistent volumes](https://mesosphere.github.io/marathon/docs/persistent-volumes.html).

(Not that it strictly matters for this PR, although I'd have preferred a formal definition. I'd file an upstream issue though.)

docker.go
```diff
@@ -46,9 +46,48 @@ type Parameters struct {
...
+
+// SetType sets the type of the PersistentVolume.
+//		t: the type of volume ("root", "path" or "mount")
+func (p *PersistentVolume) SetType(t string) *PersistentVolume {
```
timoreimann: Can be removed if we depointerize `PersistentVolume.Type`.

docker_test.go
```diff
@@ -83,3 +83,33 @@ func TestPortMappingLabels(t *testing.T) {
...
+		ContainerPath: "/foo",
+		HostPath:      "/bar",
+		Mode:          "RW",
+		Persistent:    pvol,
+	}
+
+	docker.Container.AddVolume(vol)
+
+	assert.Equal(t, 1, len(*docker.Container.Volumes))
+	assert.Equal(t, "/foo", (*docker.Container.Volumes)[0].ContainerPath)
```
timoreimann: Let's extract the first volume into a variable here and run our assertions against it.

taotetek:
Hello @timoreimann ! Thanks for the feedback!  Will address the requested changes, cheers!

taotetek:
Hello @timoreimann ! Just a heads up I haven't forgotten about this - we're in the middle of launching our mesos cluster into production so just been busy - will come back to it!

timoreimann:
Whenever you have time, @taotetek. ☺️

Thanks for the update. Good luck with the production launch! 🚀

timoreimann:
Hi @taotetek! Just wondering whether you have had a chance already to bring the PR into a mergeable state?

Thanks!

ondrej-smola:
@taotetek @timoreimann we need this. I will try to get it into mergeable state.

timoreimann:
Resolved by #295.

Thanks to both @taotetek and @ondrej-smola, I appreciate your work on this one.

PR #201, opened 2016-09-23
Expose 'Subscribe' method for event callbacks

mdirkse:
This pull request exposes a `Subscribe` function in the client so that, together with the already exposed `Unsubscribe` function, users can control event subscriptions directly, without the library starting a HTTP server to receive events.

My arguments for this change are the following:

1. I use this library from within a Docker container that runs on [Marathon](https://mesosphere.github.io/marathon/) which does port mapping from known container ports to a random machine port (which are then usually discovered via service discovery). You don't know in advance what machine port you'll get, so there is no way to reliably start a local HTTP server such that the outside world can actually communicate with it (well, it could technically be done, but it involves a lot of Marathon-specific trickery).
2. Besides, I use this library from within a http-based microservice, so I'm _already_ running a HTTP server of my own, with my own events endpoint, my own metrics, logging etc. I don't need the library to start another one for me, I just need to be able to tell Marathon the URL to which it can send events.
3. And the final point: there is currently no way for a user to register an event subscriber outside of their own app (ie outside of the process in which the library is used). Note that this isn't my use case, but could be useful to others.

All of these problems are remedied by simply exposing a `Subscribe` function in the client.


timoreimann:
@mdirkse I'd be happy to review your PR!

One question though: Have you considered using SSE-based streaming for your Marathon event needs? To my knowledge, it manages to do without a separate listener resource.

mdirkse:
@timoreimann thanks!

I did consider the SSE-based events but we're stuck on an older version of Marathon that doesn't support that, and we don't want to upgrade at the moment. Also, it doesn't address reason nr 3.

AUTHORS
```diff
@@ -34,3 +34,4 @@ Yang Bai <hamo.by@gmail.com>
 Yifa Zhang <zyfdegg@gmail.com>
 Zhanpeng Chen <lowstz@gmail.com>
 Brian Knox <taotetek@gmail.com>
+Maarten Dirkse <mdirkse@bol.com>
```
timoreimann: Nit-pick: Should ideally be sorted alphabetically. (Your predecessor is out-of-order too.)

README.md
```diff
@@ -148,6 +148,8 @@ setting with the following possible values: `EventsTransportSSE` and `EventsTran
 See [Event Stream](https://mesosphere.github.io/marathon/docs/rest-api.html#event-stream) and
 [Event Subscriptions](https://mesosphere.github.io/marathon/docs/rest-api.html#event-subscriptions) for details.
 
+Event subscriptions can also be individually controlled with the `Subscribe` and `Unsubscribe` functions.
```
timoreimann: Could you refer to the `Controlling subscriptions` section that you added below from here?

subscription.go
```diff
@@ -193,7 +192,15 @@ func (r *marathonClient) registerSSESubscription() error {
 	return nil
 }
 
-// Unsubscribe removes ourselves from Marathon's callback facility
+// Subscribe adds a URL to Marathon's callback facility
+//	url		: the url you wish to subscribe
```
timoreimann: Name of the variable is actually `callbackURL`.

Should also say `URL` in the description.

subscription.go
```diff
@@ -193,7 +192,15 @@ func (r *marathonClient) registerSSESubscription() error {
...
+func (r *marathonClient) Subscribe(callbackURL string) error {
+	uri := fmt.Sprintf("%s?callbackUrl=%s", marathonAPISubscription, callbackURL)
+	return r.apiPost(uri, "", nil)
+
+}
+
+// Unsubscribe removes a URL from Marathon's callback facility
 //	url		: the url you wish to unsubscribe
```
timoreimann: Name of the variable is actually `callbackURL`.

Should also say `URL` in the description.

subscription_test.go
```diff
@@ -178,7 +194,7 @@ func TestEventStreamEventsReceived(t *testing.T) {
 	defer endpoint.Close()
 
 	events := make(EventsChannel)
-	err := endpoint.Client.AddEventsListener(events, EventIDApplications|EventIDDeploymentInfo)
+	err := endpoint.Client.AddEventsListener(events, EventIDApplications | EventIDDeploymentInfo)
```
timoreimann: This is not gofmt-conform. Please undo or run gofmt directly.

timoreimann:
If you don't mind, could you also extend the `CHANGELOG.md` file? Not a must, just nice to have. :-)

mdirkse:
Done! Thanks for the review.

mdirkse:
My pleasure, thanks for merging :)

PR #204, opened 2016-09-24
Switch Travis CI to container-based infrastructure.

timoreimann:
Brings significant build speed improvements.


timoreimann:
Just realized there's no need to switch explicitly: Newer projects already run inside containers by default.

Closing.

PR #205, opened 2016-09-25
Signal goroutine termination on event listener removal.

### LINKED ISSUE #198: Race condition when sending events to listeners
Some background: I modified Traefik, which uses this library, to periodically shutdown the Marathon client and create a new one.  This is due to a bug in Marathon where Traefik will stop receiving events when an instance crashes.  AFAIK, this hasn't been fixed yet and I don't know when, hence the work-around.

So, after getting this code working better, I started seeing panic from the below code where it tries to send on a closed channel:

    for channel, filter := range r.listeners {
        // step: check if this listener wants this event type
        if event.ID&filter != 0 {
            go func(ch EventsChannel, e *Event) {
                ch <- e
            }(channel, event)
        }
    }

The code I'm working on when it is told to stop will remove the listener, then close the channel it was using for updates.  Looking at your code, this would appear to work fine until I realized that because you're sending the events to the channel within a goroutine, that what must be happening is the following:

1. An event comes in and a goroutine is spawned to send the event to a channel.
  * The goroutine does not run at this time.
1. My code gets a stop message.
1. The listener is removed.
1. The channel is closed.
1. The goroutine in step one now runs and panics.

Now, I do understand why the code is designed this way: you don't want to block other listeners from receiving events if one channel is full which will force the goroutine to block.  However, I'm thinking it may be best to not use a goroutine in this case and document that buffered channels should be used instead.  Perhaps a function could be added to create the correct channel vs. relying  on the user of the API to make sure it's buffered?

I understand this may be an atypical use case, but I wanted to report this bug so that it was known before this affects production systems.

timoreimann:
Fixes goroutine/memory leakage where an event consumer would stop receiving events from the channel while pending goroutines were still trying to deliver some.

The chosen approach only makes sense with unbuffered channels due to the `select` involved; as such, we change `AddEventListener`'s signature to return a channel to the consumer instead of expecting one. Passing buffered channels to `AddEventListener` was never overly useful in the
first place, however, as we use goroutines to avoid blocking on slow receivers, implictly doing event buffering this way already.

Changes in detail:

- Modify `AddEventListener`'s signature.
- Change listener's value type to struct holding both filters and a per-listener done channel.
- Close done channel on `RemoveEventListener`, and select between outbound and done channel in event-processing goroutine.
- Extend tests.
- Make existing tests more robust.
- Update examples.
- Update README file.

Fixes #198.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 - [#201][PR201]: Subscribe method is now exposed on the client to allow subscription of callback URL's
 
+### Fixed
+- [#205][PR205]: Fix memory leak by signalling goroutine termination on event listener removal.
+
+### Changed
+- [#205][PR205]: Change AddEventsListener to return event channel instead of taking one.
+
 ## [0.2.0] - 2016-09-23
+### Added
 - [#196][PR196]: Port definitions.
 - [#191][PR191]: name and labels to portMappings.
 
@@ -64,6 +71,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 [0.1.1]: https://github.com/gambol99/go-marathon/compare/v0.1.0...v0.1.1
 [0.1.0]: https://github.com/gambol99/go-marathon/compare/v0.0.1...v0.1.0
 
+[PR205]: https://github.com/gambol99/go-marathon/pull/205
 [PR202]: https://github.com/gambol99/go-marathon/pull/202
 [PR201]: https://github.com/gambol99/go-marathon/pull/201
 [PR196]: https://github.com/gambol99/go-marathon/pull/196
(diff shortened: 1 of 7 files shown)


timoreimann:
I initially planned to let the goroutine close the channel upon reception of the done signal. Unfortunately, this can still lead to send-after-close-panics if there's still a receiver listening and other goroutines pending. This is one reason why I decided to leave it out; the other is that even if it worked, the close wouldn't fire if there were no more events coming in. In summary, we'd need a more sophisticated solution to enable a correct implementation of event-listener-removal-causes-event-channel-closure, something that can be taken care of by a different PR.

@sybrandy Would you mind taking a look at let me know what you think?

@gambol99 PTAL. :-)

sybrandy:
Damn...sorry I missed this.  I lost the notification.  Regardless, it looks good to me.  Thanks!

timoreimann:
@sybrandy no worries -- thanks for taking a look!

PR #215, opened 2016-10-10
Fix unmarshalling of DeploymentPlan for Marathon v1.X API

Blahhhhh:
In [marathon docs](https://github.com/mesosphere/marathon/blob/master/docs/docs/event-bus.md), a deployment plan message is:
```
{
  ...
  "plan": {
    ...
    "steps": [
      {
        "action": "ScaleApplication",
        "app": "/my-app"
      }
    ],
    "version": "2014-03-01T23:24:14.846Z"
  },
  "currentStep": {
    "actions": [
      {
        "type": "ScaleApplication",
        "app": "/my-app"
      }
    ]
  }
}
```

But in my marathon v1.1.1 environment, the deployment plan message has got the following structure:
```
{
  ...
  "plan": {
    ...
    "steps": [
      {
        "actions": [
          {
            "type": "ScaleApplication",
            "app": "/my-app"
          }
        ]
      }
    ],
    "version": "2016-09-23T03:22:17.119Z"
  },
  "currentStep": {
    "actions": [
      {
        "type": "ScaleApplication",
        "app": "/my-app"
      }
    ]
  },
}
```

Actually, change `Steps    []*DeploymentStep` to `Steps    []*StepActions` in `DeploymentPlan` could solve the problem nice and easy, but for compatibility, I have to use this less-elegant way. 

Thanks.

### DIFF
--- a/deployment.go
+++ b/deployment.go
@@ -60,47 +60,7 @@ type DeploymentPlan struct {
 	Version  	string            	`json:"version"`
 	Original 	*Group            	`json:"original"`
 	Target   	*Group            	`json:"target"`
-	Steps    	[]*DeploymentStep 	`json:"-"`
-	XXStepsRaw	json.RawMessage		`json:"steps"` // Holds raw steps JSON to unmarshal later
-}
-
-// Allows loading of deployment steps from the Marathon v1.X API
-func (d *DeploymentPlan) UnmarshalJSON(b []byte) error {
-	dp := &struct {
-		ID       	string            	`json:"id"`
-		Version  	string            	`json:"version"`
-		Original 	*Group            	`json:"original"`
-		Target   	*Group            	`json:"target"`
-		XXStepsRaw  json.RawMessage 	`json:"steps"`
-	}{}
-	err := json.Unmarshal(b, &dp)
-	if err != nil {
-		return err
-	}
-	d.ID = dp.ID
-	d.Version = dp.Version
-	d.Original = dp.Original
-	d.Target = dp.Target
-	var deploymentStepArray []*DeploymentStep
-	if len(dp.XXStepsRaw) != 0 {
-		err = json.Unmarshal(dp.XXStepsRaw, &deploymentStepArray)
-		if err != nil {
-			return err
-		}
-		d.Steps = deploymentStepArray
-	}
-	if len(d.Steps)>0 && d.Steps[0].Action == "" {
-		var stepActionsArray []*StepActions
-		err = json.Unmarshal(dp.XXStepsRaw, &stepActionsArray)
-		if err != nil {
-			return err
-		}
-		for stepIndex, step := range stepActionsArray {
-			d.Steps[stepIndex].Action = step.Actions[0].Type
-			d.Steps[stepIndex].App = step.Actions[0].App
-		}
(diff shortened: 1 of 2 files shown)


timoreimann:
We had some prior art in this area:

- #153 (PR #158)
- #175 (PR #176)

I'll have a look at your PR.

timoreimann:
@Blahhhhh do you possibly know which version introduced this change? The older it is, the more I'd be willing to declare this is as a breaking change, forsake older Marathon users, and give up on backwards compatibility for the case at hand.

(Conservatively speaking, I'd say that any single-digit minor Marathon version is old enough to be abandoned.)

subscription_test.go
```diff
@@ -146,7 +157,77 @@ var testCases = testCaseList{
...
+				}{
+					{
+						Type: "ScaleApplication",
+						App:  "/my-app",
+					},
+				},
+			},
+		},
+	},
+	// For marathon 1.X
```
timoreimann: Can we have a similar comment for the older format that you added a test for above?

deployment.go
```diff
@@ -56,11 +56,51 @@ type StepActions struct {
...
+	err := json.Unmarshal(b, &dp)
+	if err != nil {
+		return err
+	}
+	d.ID = dp.ID
+	d.Version = dp.Version
+	d.Original = dp.Original
+	d.Target = dp.Target
+	var deploymentStepArray []*DeploymentStep
+	if len(dp.XXStepsRaw) != 0 {
```
timoreimann: There's possibly some overlap with the [implementation of Deployments()](https://github.com/gambol99/go-marathon/blob/8908223430b37fc4106d8b8cf1f38cf7124020f1/deployment.go#L67). Would be interesting to see if we can extract the logic into a common place if that doesn't make it more complicated.

Blahhhhh:
Thanks @timoreimann for reviewing. Sorry but I've got v1.1 at hand only. I'll check marathon source code to find the exact version when it changed. I also think that keeping compatibility with all marathon versions is annoying, a little change in API would lead to lots of mess code logic like this.

timoreimann:
@Blahhhhh no worries: We've got something like https://github.com/bobrik/mesos-compose at my work place. I can check a few older versions of Marathon fairly easily. Will report back once I find the time.

I agree that the compatibility code in question is nasty, and I'd love to get rid of it. It's a trade off though -- I'm fairly relaxed to abandon Marathon 0.5, but would be hesitant with 0.15 which is just one version behind then one you have reported the issue for. My hope is that Marathon 1.0 will bring more stability into the Marathon project and allow us to get rid of ugly compatibility layers over time.

timoreimann:
Sorry for the awful long delay.

I was able to verify that at least up to and include Marathon 0.10.1, the new format is already being used. I think it's okay to make this PR a breaking change since versions older than that are probably not worth supporting anymore.

Would be happy to continue the review. Please address the comments I have placed. Thanks!

Blahhhhh:
Great. If so, we can simply change a field in `DeploymentPlan` to fix the unmarshalling problem. Please review if it's ok.

timoreimann:
LGTM -- thanks for the contribution.

PR #216, opened 2016-10-10
Don't multiply duration by Milliseconds

kesselborn:
When converting pollingWaitTime from 'int' to 'time.Duration', the
default values were set to milliseconds while before they were set to
nano seconds, polling delays are 1000000 longer now which gets
problematic when using the default config

@timoreimann

### DIFF
--- a/application_test.go
+++ b/application_test.go
@@ -485,7 +485,7 @@ func TestApplicationConfiguration(t *testing.T) {
 }
 
 func TestWaitOnApplication(t *testing.T) {
-	waitTime := 100 * time.Millisecond
+	waitTime := 100 * time.Nanosecond
 
 	tests := []struct {
 		desc    string
--- a/application_test.go
+++ b/application_test.go
@@ -485,7 +485,7 @@ func TestApplicationConfiguration(t *testing.T) {
 }
 
 func TestWaitOnApplication(t *testing.T) {
-	waitTime := 100 * time.Nanosecond
+	waitTime := 100 * time.Millisecond
 
 	tests := []struct {
 		desc    string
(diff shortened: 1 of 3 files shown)


timoreimann:
@kesselborn thanks for catching, that was somewhat inadvertent. The PR that introduced the `int`-typed polling time [originally intended to operate on milliseconds scale though](https://github.com/gambol99/go-marathon/blob/ae3819f608bd44290ff53dfc93f6a7d23444ad05/config.go#L50). And before that, it was [hardcoded to seconds scale](https://github.com/gambol99/go-marathon/pull/202/files#diff-4b475da4d1c32296f086ca8abdb29c83L154). I'm not sure whether it makes sense to poll the Marathon API for updates every 500 nanoseconds by default -- that could mean a lot of traffic given the nature of the fairly slow deployment operation in Marathon.

IMHO, 500 milliseconds might be a better default for the majority of environments. WDYT?

kesselborn:
As you describe it it makes a lot of sense -- however: we ran into problems with this as if this variable was somewhere used for multiplication. I'll dive deeper into the code tomorrow and report back to see what and where the problem really caused so much trouble. Just ignore this for now until I come back with a better analysis.

kesselborn:
Ok: I found the real problem (last commit): originally, an `int` value was passed in with the notion of being milliseconds, i.e. it was multiplied by `time.Millisecond`. Now, that the type is not `int` but `time.Duration`, this mulitplication shouldn't happen.

timoreimann:
@kesselborn There was another report on the issue just today: #217. I intend to push a fix today.

PR #218, opened 2016-10-11
Remove TimeWaitPolling from marathonClient.

### LINKED ISSUE #217: If using default config, WaitOnDeployment will wait 138 hours in for loop
If I using default config, like [example/application.go](https://github.com/gambol99/go-marathon/blob/master/examples/applications/main.go#L45), it will wait for too long in [waitOnDeployment](https://github.com/gambol99/go-marathon/blob/master/examples/applications/main.go#L106) and got stuck. In fact, it will wait 138h53m20s.

After debugging, we found this issue is caused by [a mistake here](https://github.com/gambol99/go-marathon/blob/master/deployment.go#L154), I think the highlighted line should be

```go
time.Sleep(r.config.PollingWaitTime)
```

In this way, If I set timeout to 800ms like, 

```go
config := marathon.NewDefaultConfig()
config.PollingWaitTime = 800 * time.Millisecond
```

it won't be multiply by 1000000000 [here](https://github.com/gambol99/go-marathon/blob/master/client.go#L215).

WDYT? Please check and confirm this issue. I will push a commit if needed.

timoreimann:
Fixes #217.
Closes #216.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [#211][PR211] Close event channel on event listener removal.
 
 ### Fixed
+- [#218][PR218] Remove TimeWaitPolling from marathonClient.
 - [#214][PR214] Remove extra pointer layers when passing to r.api*.
 
 ## [0.3.0] - 2016-09-28
@@ -79,6 +80,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 [0.1.1]: https://github.com/gambol99/go-marathon/compare/v0.1.0...v0.1.1
 [0.1.0]: https://github.com/gambol99/go-marathon/compare/v0.0.1...v0.1.0
 
+[PR218]: https://github.com/gambol99/go-marathon/pull/218
 [PR214]: https://github.com/gambol99/go-marathon/pull/214
 [PR211]: https://github.com/gambol99/go-marathon/pull/211
 [PR205]: https://github.com/gambol99/go-marathon/pull/205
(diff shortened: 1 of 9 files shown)


kesselborn:
haha ... yeah: even better

timoreimann:
@kesselborn @zyfdegh I pushed one last commit (3f77d47). Would you mind taking a look? See also the commit message for a rationale.

zyfdegh:
Good job. It looks better now. But why not use

```go
conf := cl.config
```

directly instead of

```sh
conf := cl.(*marathonClient).conf()
```

inside `func TestNewClient(t *testing.T)`.

I'm a bit confused. What will happen if call directly?

timoreimann:
@zyfdegh The configuration instance is not exported by `marathonClient`, so we can only access it local to the client. I thought about changing its accessibility but decided against it as I don't want to give the impression that the config is something that can (or should) be changed after the client has been created. For similar reasons, I wouldn't want to turn `config` into a pointer (plus, it'd be breaking change).

The type assertion is a bit ugly and will panic if we, for some reason, cast the wrong instance. However, given  that we should be calling this method from a test context only, it's acceptable to me.

Does that sound reasonable?

timoreimann:
Geez, I think what I just wrote actually doesn't make sense as the function I created and the conf struct live on the same accessibility space. Let me double check and simplify if I can. :-)

timoreimann:
@zyfdegh I was right and wrong :-)

Even though the export status isn't a problem, `NewClient` returns an interface which obviously cannot hold data. Thus, we can get rid of the test helper function but not the type assertion.

I pushed 811fb08.

timoreimann:
Time to merge.

zyfdegh:
OK, got it. Anyway, it works good for me.

PR #220, opened 2016-10-12
Add support for external volumes

bosschaert:
External volumes allow integration with persistent volumes such as NFS or EBS volumes that can be attached to the nodes running the containers. This allows the persistent data to outlive the container in case of a crash and makes it possible to migrate it to wherever the container is moving to.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -46,9 +46,17 @@ type Parameters struct {
 
 // Volume is the docker volume details associated to the container
 type Volume struct {
-	ContainerPath string `json:"containerPath,omitempty"`
-	HostPath      string `json:"hostPath,omitempty"`
-	Mode          string `json:"mode,omitempty"`
+	ContainerPath  string          `json:"containerPath,omitempty"`
+	HostPath       string          `json:"hostPath,omitempty"`
+	ExternalVolume *ExternalVolume `json:"external,omitempty"`
+	Mode           string          `json:"mode,omitempty"`
+}
+
+// An external volume definition
+type ExternalVolume struct {
+	Name     string             `json:"name,omitempty"`
+	Provider string             `json:"provider,omitempty"`
+	Options  *map[string]string `json:"options,omitempty"`
 }
 
 // Docker is the docker definition from a marathon application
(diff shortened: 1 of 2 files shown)


bosschaert:
I just noticed that #200 aims to provide similar functionality. Hope the best implementation can be merged :)

bosschaert:
FWIW the 'merge branch 'master' into external-volumes' was created when I clicked the 'update' button in github. If you prefer I can rebase my commit so that it's a single commit that can be fast-forwarded.

timoreimann:
Already started reviewing. Will send out a few comments soon. ☺️

docker.go
```diff
@@ -46,9 +46,17 @@ type Parameters struct {
...
+	ContainerPath  string          `json:"containerPath,omitempty"`
+	HostPath       string          `json:"hostPath,omitempty"`
+	ExternalVolume *ExternalVolume `json:"external,omitempty"`
+	Mode           string          `json:"mode,omitempty"`
+}
+
+// An external volume definition
+type ExternalVolume struct {
+	Name     string             `json:"name,omitempty"`
+	Provider string             `json:"provider,omitempty"`
```
timoreimann: I suppose neither `Name` nor `Provider` may be empty strings from an API point of view?

In case they may, we need to pointerize the strings so that `omitempty` does not drop them.

docker.go
```diff
@@ -90,6 +98,36 @@ func (container *Container) EmptyVolumes() *Container {
 	return container
 }
 
+// Define external volumes attached to the container
+//      containerPath: the path inside the container to map the volume
+//      name: the name of the volume
+//      provider: the provider of the volume (e.g. dvdi)
+//      mode: the read/write mode for the volume (e.g. RW)
+//      opts: the map of options for the external volume
+func (container *Container) ExternalVolume(containerPath string,
```
timoreimann: - This duplicates a bit of functionality from [Volume()](https://github.com/bosschaert/go-marathon/blob/18b68aa3267107dc7ea9e6bbfc3f77b659273906/docker.go#L76). I think `ExternalVolume` should only receive parameters specific to an external volume (i.e., `name`, `provider`, and `opts`) and just create an `ExternalVolume`. Users would then call the method, assign the result to a `Volume` instance, and add that volume to the container.
- You can reduce the consecutive string parameter declarations by doing `name, provider, mode string`.
- `opts` is a bit unwieldy if you don't need it as you'd have to pass in an empty map like `map[string]string{}`. Plus, having this parameter conceptually means we'd need to extend the signature in a breaking way for every new parameter we get in the future. How about attaching a method `SetOpts()` to `ExternalVolume` that takes a map instead? If we pair it with an `EmptyOpts()` method, we also allow users to clear options.

docker_test.go
```diff
@@ -83,3 +83,31 @@ func TestPortMappingLabels(t *testing.T) {
 	assert.NotNil(t, pm.Labels)
 	assert.Equal(t, 0, len(*pm.Labels))
 }
+
+func TestSimpleVolumes(t *testing.T) {
```
timoreimann: I think we can drop this one if we decide to narrow `ExternalVolume`'s purpose as suggested elsewhere.

bosschaert:
Thanks for the comments @timoreimann - I'll update the branch some time soon to hopefully address your concerns.

docker.go
bosschaert: Looking at the json schema in http://mesosphere.github.io/marathon/docs/generated/api.html#v2_apps__app_id__put it seems like Name and Provider are, at least from an API definition point of view, not required to have a value. 
So I'm inclined to leave these the way they are. In my use-case they definitely always have a value, but there might be some usage out there that doesn't require this.

bosschaert:
I have refactored and rebased the code and created a new pull request at #221. So I'm closing this one.

timoreimann:
@bosschaert would not have been necessary to close the PR and create a new one. Happy to review the new PR though. :-)

PR #221, opened 2016-10-13
Add external volume definition to container

bosschaert:
External volumes allow integration with persistent volumes such as NFS or EBS volumes that can be attached to the nodes running the containers. This allows the persistent data to outlive the container in case of a crash and makes it possible to migrate it to wherever the container is moving to.

This pull request replaces #220

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -46,9 +46,17 @@ type Parameters struct {
 
 // Volume is the docker volume details associated to the container
 type Volume struct {
-	ContainerPath string `json:"containerPath,omitempty"`
-	HostPath      string `json:"hostPath,omitempty"`
-	Mode          string `json:"mode,omitempty"`
+	ContainerPath string          `json:"containerPath,omitempty"`
+	HostPath      string          `json:"hostPath,omitempty"`
+	External      *ExternalVolume `json:"external,omitempty"`
+	Mode          string          `json:"mode,omitempty"`
+}
+
+// An external volume definition
+type ExternalVolume struct {
+	Name     string             `json:"name,omitempty"`
+	Provider string             `json:"provider,omitempty"`
+	Options  *map[string]string `json:"options,omitempty"`
 }
 
 // Docker is the docker definition from a marathon application
@@ -88,6 +96,37 @@ func (container *Container) Volume(hostPath, containerPath, mode string) *Contai
 func (container *Container) EmptyVolumes() *Container {
 	container.Volumes = &[]Volume{}
 	return container
+}
+
+// Define external elements for a volume
+//      name: the name of the volume
+//      provider: the provider of the volume (e.g. dvdi)
+func (v *Volume) ExternalVolume(name, provider string) *ExternalVolume {
+	ev := ExternalVolume{
+		Name:     name,
+		Provider: provider,
+	}
+	v.External = &ev
+	return &ev
+}
+
+// AddOption adds an option to an ExternalVolume
+//		name:  the name of the option
+//		value: value for the option
+func (ev *ExternalVolume) AddOption(name, value string) *ExternalVolume {
+	if ev.Options == nil {
+		ev.EmptyOptions()
+	}
+	(*ev.Options)[name] = value
+
+	return ev
+}
+
+// EmptyOptions explicitly empties the options
+func (ev *ExternalVolume) EmptyOptions() *ExternalVolume {
+	ev.Options = &map[string]string{}
+
+	return ev
 }
 
 // NewDockerContainer creates a default docker container for you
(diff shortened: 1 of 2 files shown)


bosschaert:
@timoreimann hopefully this pull request addresses your concerns.

docker.go
```diff
@@ -90,6 +98,37 @@ func (container *Container) EmptyVolumes() *Container {
 	return container
 }
 
+// Define external elements for a volume
+//      name: the name of the volume
+//      provider: the provider of the volume (e.g. dvdi)
+func (v *Volume) ExternalVolume(name, provider string) *ExternalVolume {
+	ev := ExternalVolume{
```
timoreimann: nit-pick: You could take the address of `ExternalVolume` directly on the assignment like

```go
ev := &ExternalVolume{
  ...
}
```

and save the address operators below (net win: one `&` less :-) ).

docker_test.go
```diff
@@ -83,3 +83,23 @@ func TestPortMappingLabels(t *testing.T) {
 	assert.NotNil(t, pm.Labels)
 	assert.Equal(t, 0, len(*pm.Labels))
 }
+
+func TestExternalVolumes(t *testing.T) {
```
timoreimann: Please rename to `TestExternalVolume` (singular volume).

docker_test.go
```diff
@@ -83,3 +83,23 @@ func TestPortMappingLabels(t *testing.T) {
...
+	container := NewDockerApplication().Container
+
+	container.Volume("", "cp", "RW")
+	ev := (*container.Volumes)[0].ExternalVolume("myVolume", "dvdi")
+
+	ev.AddOption("prop", "pval")
+	ev.AddOption("dvdi", "rexray")
+
+	assert.Equal(t, (*container.Volumes)[0].ContainerPath, "cp")
```
timoreimann: The test should not need to care about fields unrelated to `ExternalVolume`, so I'd drop the check on `ContainerPath`.

docker_test.go
```diff
@@ -83,3 +83,23 @@ func TestPortMappingLabels(t *testing.T) {
...
+	assert.Equal(t, (*container.Volumes)[0].External.Name, "myVolume")
+	assert.Equal(t, (*container.Volumes)[0].External.Provider, "dvdi")
+	assert.Equal(t, 1, len(*container.Volumes))
```
timoreimann: This belongs to a test dedicated to validate that `Volume()` works correctly. Unfortunately, we didn't implement such a test when we added `Volume()`. :-(

I'll leave it up to you to either create `TestVolume` and move the length check there, or just drop it in your test.

docker_test.go
```diff
@@ -83,3 +83,23 @@ func TestPortMappingLabels(t *testing.T) {
...
+
+	evt := (*container.Volumes)[0].External
+	assert.Equal(t, (*evt.Options)["dvdi"], "rexray")
+	assert.Equal(t, (*evt.Options)["prop"], "pval")
+	assert.Equal(t, 2, len(*evt.Options))
```
timoreimann: Please move the length check right after the definition of `evt`. You should probably also let the other two option tests depend on the result of the length check, as in:

```go
evt := (*container.Volumes)[0].External
if assert.Equal(t, 2, len(*evt.Options)) {
  assert.Equal(t, (*evt.Options)["dvdi"], "rexray")
  assert.Equal(t, (*evt.Options)["prop"], "pval")
}
```

docker.go
```diff
@@ -90,6 +98,37 @@ func (container *Container) EmptyVolumes() *Container {
...
+		Name:     name,
+		Provider: provider,
+	}
+	v.External = &ev
+	return &ev
+}
+
```
timoreimann: Would be great if we could also have an `EmptyVolume()` method. `TestExternalVolume()` should then be extended by a call to that function and a check that the external volume is actually reset.

timoreimann:
Another smallish review round. :-)

bosschaert:
@timoreimann thanks for the review and feedback. Hopefully I addressed your concerns in this commit.

docker.go
```diff
@@ -90,6 +98,42 @@ func (container *Container) EmptyVolumes() *Container {
 	return container
 }
 
+// Define external elements for a volume
+//      name: the name of the volume
+//      provider: the provider of the volume (e.g. dvdi)
+func (v *Volume) ExternalVolume(name, provider string) *ExternalVolume {
```
timoreimann: Can we call this `SetExternalVolume`?

docker.go
```diff
@@ -90,6 +98,42 @@ func (container *Container) EmptyVolumes() *Container {
...
+	ev := &ExternalVolume{
+		Name:     name,
+		Provider: provider,
+	}
+	v.External = ev
+	return ev
+}
+
+func (v *Volume) EmptyVolume() *Volume {
```
timoreimann: Should be `EmptyExternalVolume`.

bosschaert:
Thanks @timoreimann I really appreciate your help and time! I did the renames in this commit.

timoreimann:
Thanks for your contribution -- merged!

PR #222, opened 2016-10-20
make allHealthChecksAlive public so as a user I can test if task passed

cmingxu:
alive check

Signed-off-by: kevin xu <cming.xu@gmail.com>

### DIFF
--- a/task.go
+++ b/task.go
@@ -199,7 +199,7 @@ func (r *marathonClient) TaskEndpoints(name string, port int, healthCheck bool) 
 	// step: iterate the tasks and extract the dynamic ports
 	var list []string
 	for _, task := range application.Tasks {
-		if !healthCheck || task.allHealthChecksAlive() {
+		if !healthCheck || task.AllHealthChecksAlive() {
 			endpoint := fmt.Sprintf("%s:%d", task.Host, task.Ports[portIndex])
 			list = append(list, endpoint)
 		}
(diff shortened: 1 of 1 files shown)


timoreimann:
@cmingxu out of curiosity, does `ApplicationOK` not work for you?

cmingxu:
@timoreimann Thanks for the quick response, I want expose individual task to the LB at the front of a group of tasks.
I suppose ApplicationOK return the status of the App but for a instance should at task level.

cmingxu:
@timoreimann Thanks for the quick response, I want expose individual task to the LB at the front of a group of tasks.
I suppose ApplicationOK return the status of the App but for a instance should at task level.

timoreimann:
True, `ApplicationOK` looks at the set of tasks for a specific app only. I see you want more flexibility.

Will follow up with a review.

task.go
```diff
@@ -208,7 +208,7 @@ func (r *marathonClient) TaskEndpoints(name string, port int, healthCheck bool)
 	return list, nil
 }
 
-func (r *Task) allHealthChecksAlive() bool {
+func (r *Task) AllHealthChecksAlive() bool {
```
timoreimann: The implementation assumes that we are dealing with a task configured with custom health checks. This means that users need to make sure they are calling the method on proper tasks only (like the single caller in go-marathon does) to avoid returning a negative result for an otherwise healthy task.

I see two options to mitigate this problem:

1. Document the circumstance in the method's godoc.
1. Create a new method on `Application` that returns a list of healthy tasks, taking any configured health checks into account.

I'm leaning towards 2., not sure though if that fits your use case. Any thoughts / preferences?

cmingxu: @timoreimann , I prefer option 1 , leave it to people who use this method, as if health check differ from task to task, people who set it should know.

timoreimann: Fine with me. We should then

- comment the newly exported method, explaining when it returns true and false, respectively; and 
- make sure we cover all code paths through a test. (Can't do this right now while being on the phone, unfortunately.)

timoreimann:
?

PR #223, opened 2016-10-25
Adding support for ip per task

diegooliveira:
Structs, docs and tests.

 The last PR to do this was a little old and missing helper methods and tests.

### DIFF
--- a/application.go
+++ b/application.go
@@ -34,6 +34,25 @@ type Applications struct {
 	Apps []Application `json:"apps"`
 }
 
+// IPAddressPerTask is used by IP-per-task functionality https://mesosphere.github.io/marathon/docs/ip-per-task.html
+type IPAddressPerTask struct {
+	Groups    *[]string          `json:"groups,omitempty"`
+	Labels    *map[string]string `json:"labels,omitempty"`
+	Discovery *Discovery         `json:"discovery,omitempty"`
+}
+
+// Discovery provides info about ports expose by Ip-per-task functionality
+type Discovery struct {
+	Ports *[]Port `json:"ports,ommitempty"`
+}
+
+// Port provides info about ports used by Ip-per-task
+type Port struct {
+	Number   int    `json:"number,omitempty"`
+	Name     string `json:"name,omitempty"`
+	Protocol string `json:"protocol,omitempty"`
+}
+
 // Application is the definition for an application in marathon
 type Application struct {
 	ID                    string              `json:"id,omitempty"`
@@ -70,6 +89,7 @@ type Application struct {
 	AcceptedResourceRoles []string            `json:"acceptedResourceRoles,omitempty"`
 	LastTaskFailure       *LastTaskFailure    `json:"lastTaskFailure,omitempty"`
 	Fetch                 *[]Fetch            `json:"fetch,omitempty"`
+	IPAddressPerTask      *IPAddressPerTask   `json:"ipAddress,omitempty"`
 }
 
 // ApplicationVersions is a collection of application versions for a specific app in marathon
@@ -107,6 +127,28 @@ type GetAppOpts struct {
 //		force:		overrides a currently running deployment.
 type DeleteAppOpts struct {
 	Force bool `url:"force,omitempty"`
+}
+
+// Diego ======
+
+// SetIPAddressPerTask defines that the application will have a IP address defines by a external agent. this
+// configuration is not allowed to be used with Port or PortDefinitions.
+func (r *Application) SetIPAddressPerTask(ipAddressPerTask *IPAddressPerTask) *Application {
+
+	r.Ports = make([]int, 0)
+	r.EmptyPortDefinitions()
+	r.IPAddressPerTask = ipAddressPerTask
+
+	return r
+}
+
+// EmptyIPAddressPerTask explicitly empties port definitions -- use this if you need to empty
+// IP address per task of an application that already has some definitions set (setting port definitions to nil will
+// keep the current value)
+func (r *Application) EmptyIPAddressPerTask() *Application {
+	r.IPAddressPerTask = nil
+
+	return r
 }
 
 // NewDockerApplication creates a default docker application
@@ -731,4 +773,87 @@ func buildURIWithForceParam(path string, force bool) string {
 
 func buildURI(path string) string {
 	return fmt.Sprintf("%s/%s", marathonAPIApps, trimRootPath(path))
+}
+
+// EmptyLabels explicitly empties labels -- use this if you need to empty
+// labels of an application that already has ip per task with labels defined
+func (i *IPAddressPerTask) EmptyLabels() *IPAddressPerTask {
+
+	i.Labels = &map[string]string{}
+
+	return i
+}
+
+// AddLabel adds a label to an IPAddressPerTask
+//    name: The label name
+//   Value: The label value
+func (i *IPAddressPerTask) AddLabel(name, value string) *IPAddressPerTask {
+
+	if i.Labels == nil {
+		i.EmptyLabels()
+	}
+
+	(*i.Labels)[name] = value
+
+	return i
+}
+
+// EmptyGroups explicitly empties groups -- use this if you need to empty
+// groups of an application that already has ip per task with groups defined
+func (i *IPAddressPerTask) EmptyGroups() *IPAddressPerTask {
+
+	i.Groups = &[]string{}
+
+	return i
+}
+
+// AddGroup adds a group to an IPAddressPerTask
+//  group: The group name
+func (i *IPAddressPerTask) AddGroup(group string) *IPAddressPerTask {
+	if i.Groups == nil {
+		i.EmptyGroups()
+	}
+
+	groups := *i.Groups
+	groups = append(groups, group)
+	i.Groups = &groups
+
+	return i
+}
+
+// EmptyDiscovery explicitly empties discovey data -- use this if you need to empty
+// discovey of an application that already has ip per task with discovey defined
+func (i *IPAddressPerTask) EmptyDiscovery() *IPAddressPerTask {
+	i.Discovery = nil
+	return i
+}
+
+// SetDiscovery define the discovery to an IPAddressPerTask
+//  discovery: The discovery struct
+func (i *IPAddressPerTask) SetDiscovery(discovery *Discovery) *IPAddressPerTask {
+	i.Discovery = discovery
+	return i
+}
+
+// EmptyPorts explicitly empties discovey port -- use this if you need to empty
+// discovey port of an application that already has ip per task with discovey ports
+// defined
+func (d *Discovery) EmptyPorts() *Discovery {
+	d.Ports = &[]Port{}
+
+	return d
+}
+
+// AddPort adds a port to the discovery info of a ip per task applicable
+//   port: The discovery port
+func (d *Discovery) AddPort(port Port) *Discovery {
+	if d.Ports == nil {
+		d.EmptyPorts()
+	}
+
+	ports := *d.Ports
+	ports = append(ports, port)
+	d.Ports = &ports
+
+	return d
 }
--- a/application.go
+++ b/application.go
@@ -38,15 +38,15 @@ type Applications struct {
 type IPAddressPerTask struct {
 	Groups    *[]string          `json:"groups,omitempty"`
 	Labels    *map[string]string `json:"labels,omitempty"`
-	Discovery *Discovery         `json:"discovery,omitempty"`
+	Discovery Discovery          `json:"discovery,omitempty"`
 }
 
-// Discovery provides info about ports expose by Ip-per-task functionality
+// Discovery provides info about ports expose by IP-per-task functionality
 type Discovery struct {
 	Ports *[]Port `json:"ports,omitempty"`
 }
 
-// Port provides info about ports used by Ip-per-task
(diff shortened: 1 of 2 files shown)


gambol99:
LG ... @timoreimann? ...

diegooliveira:
@gambol99 done s/ommitempty/omitempty/

application.go
```diff
@@ -109,6 +129,28 @@ type DeleteAppOpts struct {
...
+	r.IPAddressPerTask = ipAddressPerTask
+
+	return r
+}
+
+// EmptyIPAddressPerTask explicitly empties port definitions -- use this if you need to empty
+// IP address per task of an application that already has some definitions set (setting port definitions to nil will
+// keep the current value)
+func (r *Application) EmptyIPAddressPerTask() *Application {
+	r.IPAddressPerTask = nil
```
timoreimann: The idea of the `Empty*` methods is that we support clearing a value previously set. However, setting it to nil would simply not pass the field in the resulting JSON due to our usage of the omitempty tag. What we would need to do instead is pass the address of an empty `IPAddressPerTask` struct, i.e., `&IPAddressPerTask{}`.

For the case at hand, however, it doesn't have the expected effect: According to the documentation, an empty `ipAddress` field yields the default configuration for an enabled IP-per-task setting. The API does not seem to provide a way to disable the feature once enabled.

If you share my observation, then I suggest we remove `EmptyIPAddressPerTask` entirely in order to avoid confusing users into believing they are getting a feature which they don't. (If someone wants to return to the default setting after having set something more specific (like discovery), they can still do so by calling `SetIPAddressPerTask` with an empty struct.)

Apologies for having mislead in the other PR by originally advocating in favor of this method. I only realized now that the field we are dealing with carries different semantics.

application.go
```diff
@@ -109,6 +129,28 @@ type DeleteAppOpts struct {
 	Force bool `url:"force,omitempty"`
 }
 
+// Diego ======
+
+// SetIPAddressPerTask defines that the application will have a IP address defines by a external agent. this
+// configuration is not allowed to be used with Port or PortDefinitions.
+func (r *Application) SetIPAddressPerTask(ipAddressPerTask *IPAddressPerTask) *Application {
+
```
timoreimann: No extra blank line.

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
 func buildURI(path string) string {
 	return fmt.Sprintf("%s/%s", marathonAPIApps, trimRootPath(path))
 }
+
+// EmptyLabels explicitly empties labels -- use this if you need to empty
+// labels of an application that already has ip per task with labels defined
+func (i *IPAddressPerTask) EmptyLabels() *IPAddressPerTask {
+
```
timoreimann: No extra blank line.

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
...
+	(*i.Labels)[name] = value
+
+	return i
+}
+
+// EmptyGroups explicitly empties groups -- use this if you need to empty
+// groups of an application that already has ip per task with groups defined
+func (i *IPAddressPerTask) EmptyGroups() *IPAddressPerTask {
+
```
timoreimann: No extra blank line.

application.go
```diff
@@ -109,6 +129,28 @@ type DeleteAppOpts struct {
 	Force bool `url:"force,omitempty"`
 }
 
+// Diego ======
+
+// SetIPAddressPerTask defines that the application will have a IP address defines by a external agent. this
+// configuration is not allowed to be used with Port or PortDefinitions.
+func (r *Application) SetIPAddressPerTask(ipAddressPerTask *IPAddressPerTask) *Application {
```
timoreimann: I'd let the method take a non-pointer. It simplifies the usage while we shouldn't need pointer semantics for performance or other reasons. Plus, we'd be consistent with the other methods added by this PR.

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
...
+	i.Labels = &map[string]string{}
+
+	return i
+}
+
+// AddLabel adds a label to an IPAddressPerTask
+//    name: The label name
+//   Value: The label value
+func (i *IPAddressPerTask) AddLabel(name, value string) *IPAddressPerTask {
+
```
timoreimann: No extra blank line.

application.go
```diff
@@ -109,6 +129,28 @@ type DeleteAppOpts struct {
 	Force bool `url:"force,omitempty"`
 }
 
+// Diego ======
+
+// SetIPAddressPerTask defines that the application will have a IP address defines by a external agent. this
+// configuration is not allowed to be used with Port or PortDefinitions.
+func (r *Application) SetIPAddressPerTask(ipAddressPerTask *IPAddressPerTask) *Application {
+
+	r.Ports = make([]int, 0)
```
timoreimann: Do we need this? To me, the documentation reads like we only need to clear port definitions. I never used the feature in question though.

application.go
```diff
@@ -34,6 +34,25 @@ type Applications struct {
...
 }
 
+// IPAddressPerTask is used by IP-per-task functionality https://mesosphere.github.io/marathon/docs/ip-per-task.html
+type IPAddressPerTask struct {
+	Groups    *[]string          `json:"groups,omitempty"`
+	Labels    *map[string]string `json:"labels,omitempty"`
+	Discovery *Discovery         `json:"discovery,omitempty"`
+}
+
+// Discovery provides info about ports expose by Ip-per-task functionality
```
timoreimann: `s/Ip/IP/`

application.go
```diff
@@ -34,6 +34,25 @@ type Applications struct {
...
+type Discovery struct {
+	Ports *[]Port `json:"ports,omitempty"`
+}
+
+// Port provides info about ports used by Ip-per-task
```
timoreimann: `s/Ip/IP/`

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
 func buildURI(path string) string {
 	return fmt.Sprintf("%s/%s", marathonAPIApps, trimRootPath(path))
 }
+
+// EmptyLabels explicitly empties labels -- use this if you need to empty
+// labels of an application that already has ip per task with labels defined
```
timoreimann: `s/ip/IP/`

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
...
+func (i *IPAddressPerTask) EmptyLabels() *IPAddressPerTask {
+
+	i.Labels = &map[string]string{}
+
+	return i
+}
+
+// AddLabel adds a label to an IPAddressPerTask
+//    name: The label name
+//   Value: The label value
```
timoreimann: `s/Value/value`

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
...
+		i.EmptyLabels()
+	}
+
+	(*i.Labels)[name] = value
+
+	return i
+}
+
+// EmptyGroups explicitly empties groups -- use this if you need to empty
+// groups of an application that already has ip per task with groups defined
```
timoreimann: `s/ip/IP/`

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
...
+
+	groups := *i.Groups
+	groups = append(groups, group)
+	i.Groups = &groups
+
+	return i
+}
+
+// EmptyDiscovery explicitly empties discovey data -- use this if you need to empty
+// discovey of an application that already has ip per task with discovey defined
```
timoreimann: `s/ip/IP/`

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
...
+func (i *IPAddressPerTask) EmptyDiscovery() *IPAddressPerTask {
+	i.Discovery = nil
```
timoreimann: This should not be nil but an empty `Discovery` struct.

I'm presuming here that the empty value clears a discovery setting. If it does not, we should drop the method as well.

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
...
+	return i
+}
+
+// SetDiscovery define the discovery to an IPAddressPerTask
+//  discovery: The discovery struct
+func (i *IPAddressPerTask) SetDiscovery(discovery *Discovery) *IPAddressPerTask {
```
timoreimann: I'd use a non-pointer argument again.

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
...
+	i.Discovery = discovery
+	return i
+}
+
+// EmptyPorts explicitly empties discovey port -- use this if you need to empty
+// discovey port of an application that already has ip per task with discovey ports
```
timoreimann: `s/ip/IP/`

application.go
```diff
@@ -732,3 +774,86 @@ func buildURIWithForceParam(path string, force bool) string {
...
+// defined
+func (d *Discovery) EmptyPorts() *Discovery {
+	d.Ports = &[]Port{}
+
+	return d
+}
+
+// AddPort adds a port to the discovery info of a ip per task applicable
```
timoreimann: `s/ip/IP/`

application_test.go
```diff
@@ -564,3 +564,39 @@ func TestAppExistAndRunning(t *testing.T) {
 	assert.True(t, client.appExistAndRunning(fakeAppName))
 	assert.False(t, client.appExistAndRunning("no_such_app"))
 }
+
+func TestSetIpPerTask(t *testing.T) {
```
timoreimann: `s/Ip/IP/`

application_test.go
```diff
@@ -564,3 +564,39 @@ func TestAppExistAndRunning(t *testing.T) {
...
+func TestDiscovery(t *testing.T) {
+	disc := Discovery{}
+	assert.Nil(t, disc.Ports)
+
+	disc.AddPort(Port{})
+	assert.NotNil(t, disc.Ports)
+
+	disc.EmptyPorts()
+	assert.Equal(t, 0, len(*disc.Ports))
+}
```
timoreimann: Please also incorporate the helper methods to set and clear groups and labels, respectively, in your tests.

timoreimann:
Bookkeeping note: old PR is at #163

application.go
diegooliveira: The API describes that using ports/ports definition is not allowed if using IP-per-task. Actualy if you send a post with both options you'll get an error. I was asking myself to put or not this cleanup, it is like reproduce an API behavior in the client library.

application.go
diegooliveira: I did some tests in this feature and it's exactly what you said, once enabled IP-per-task it is impossible to disable. I'll remove the method.

application.go
timoreimann: Personally, I usually prefer to let the API do the error handling as it means less coupling. In this very case though, I think it's a convenience add-on we can justify to keep in. (Especially since the documentation is a bit incomplete here.)

application_test.go
```diff
@@ -564,3 +564,51 @@ func TestAppExistAndRunning(t *testing.T) {
...
+
+	ipPerTask.EmptyGroups()
+	assert.Equal(t, 0, len(*ipPerTask.Groups))
+
+	ipPerTask.EmptyLabels()
+	assert.Equal(t, 0, len(*ipPerTask.Labels))
+
+}
+
+func TestDiscovery(t *testing.T) {
```
timoreimann: Can we rename this to `TestIPAddressPerTaskDiscovery`?

application.go
```diff
@@ -109,6 +129,16 @@ type DeleteAppOpts struct {
 	Force bool `url:"force,omitempty"`
 }
 
+// SetIPAddressPerTask defines that the application will have a IP address defines by a external agent.
+// This configuration is not allowed to be used with Port or PortDefinitions.
```
timoreimann: Let's amend the sentence by something like 

>,  which is why the implementation clears both.

application.go
```diff
@@ -109,6 +129,16 @@ type DeleteAppOpts struct {
...
+func (r *Application) SetIPAddressPerTask(ipAddressPerTask *IPAddressPerTask) *Application {
```
timoreimann: I'm still in favor of passing in a non-pointer to simplify things for the user.

diegooliveira: If you pass non pointer the client will always send this field in the json body, thus marathon will think it's a default IP-per-task configuration. Using it as a pointer will avoid this behavior.

timoreimann: I only meant the parameter as passed to the `SetIPAddressPerTask` method, not the field in the `Application` struct.

I'm basically in favor of making just the following change:

```go
func (r *Application) SetIPAddressPerTask(ipAddressPerTask IPAddressPerTask) *Application {
  r.Ports = make([]int, 0)
  r.EmptyPortDefinitions()
  r.IPAddressPerTask = &ipAddressPerTask
 
  return r
}
```

(Address operator gone in the method signature and "moved" to the assignment.)

It's just a tad easier for the user and indicates that we don't take `nil`s.

That I was also my point when [talking about discovery](https://github.com/gambol99/go-marathon/pull/223#discussion_r85009573). I suppose there was a misunderstanding, and that's why you de-pointerized the `Discovery` field in the `IPAddressPerTask` struct.

PR #226, opened 2016-10-26
Add Coveralls.io code coverage badge.

timoreimann:


jangie:
Out of curiosity, past changing the branch that is listed in the querystring, is there any reason this isn't being merged?

timoreimann:
It's been a while, but IIRC, I couldn't get it to work back then and ran out of time at one point.

Not sure what the current state is, to be honest. Happy to let you take over if you feel like. :-)

PR #229, opened 2016-11-03
Add LastFailureCause field to HealthCheckResult

imakewebthings:
*Problem:* Marathon health check results include a `LastFailureCause` field that describes the reason for the last health check failure, but the `HealthCheckResult` struct doesn't include a field for it.

I can't find this field explicitly documented in the Marathon API docs, but it's certainly [part of a health check result](https://github.com/mesosphere/marathon/blob/31426bd1f7b091b78748ea7fd83c497db3dfea9e/src/main/scala/mesosphere/marathon/core/health/Health.scala#L11), and in the JSON returned from the tasks endpoints. 

screen shot 2016-11-03 at 10 52 33 am

Would love to have this exposed. Thanks!

### DIFF
--- a/health.go
+++ b/health.go
@@ -82,6 +82,7 @@ type HealthCheckResult struct {
 	ConsecutiveFailures int    `json:"consecutiveFailures"`
 	FirstSuccess        string `json:"firstSuccess"`
 	LastFailure         string `json:"lastFailure"`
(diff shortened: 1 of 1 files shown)


timoreimann:
Sounds perfectly reasonable. Thanks for contributing!

PR #231, opened 2016-11-04
Marathon Cluster Markdowns

gambol99:
- removed the linked lists and moved to slices
- fixed up the code to manage the selection of a node
- @NOTE: this wouldnt fix the events subscription at the moment, need to work that in
- changed some of the names of the structs and removed the cluster interface, not sure point in it


sybrandy:
I took a quick look and like the new logic, especially healthCheckNode.  I have two things you may want to think about, though neither is a show stopper.

1. For healthCheckNode, perhaps an exponential back-off would be a good idea?  It by itself probably isn't an issue, but it may be nice to cut back on the network traffic in case it's down for a good while and other services are trying to connect to it.
2. For getMember, would it make sense to add some rudimentary load-balancing?  I'm thinking simple round-robin to start.  We have very few things hitting Marathon in our cluster, but busier clusters may have more, so balancing the load a bit is a good idea.  However, it could be argued that if you have that many apps hitting Marathon you're probably doing something wrong.

However, markDown bothers me.  I'm not sure how I can prove it, but it looks like there could be a race condition where a node that is up will be marked down by accident.  I suspect the health check will cause it to recover, but if all of the nodes are temporarily marked down, that may throw some unnecessary errors.  Below is the scenario that I think will cause this:

1. Goroutine A calls getMember (host1).
2. Goroutine B calls getMember (host1).
3. Goroutine A can't connect to the member (host1) and calls markDown (host1 now down).
4. Goroutine B can't connect to the member (host1) and calls markDown (host2 now down).
5. Goroutine A calls getMember (host3).

Hopefully this explains it.  Again, I'm not sure how to prove this off the top of my head and it's probably not a major issue, however I wanted to mention it just in case.  At this time, unless it becomes a major issue, I don't know if it's worth refactoring it to prevent the issue.

Just in case I come of as being critical, please note that I do like this change a lot.  Nothing that I mentioned should stop this from being merged in.

Thanks!

gambol99:
@sybrandy ... yep, makes sense, given there is no information being passed on the markdown() is possible to get routines which depending on where one is preempted could mark down successive nodes .. The easiest solution would be to change it to markDown(endpoint) specifying endpoint ... The backoff in the health check sounds sensible so i'll it to the PR

In regard to the load balancing, we could add another client option FAILOVER, ROUND_ROBIN? etc ...

timoreimann:
1. I also think that we should fix `markDown`, which could be as easy as passing the node/index to the function to make sure we don't run into the race.
2. Marathon followers always forward API traffic to the leader. Thus, load-balancing would only gain us something for the cases where we manage to hit the leader so that we save one hop. A better approach might be to periodically determine the leader and send traffic directly to it. I'm not sure though if there are users out there that will actually benefit from this kind of routing optimization. Plus, we'd need to make assumptions that could turn invalid in the future (like Marathon switching to a model where each node can respond to a request).
3. Exponential backoff sounds like a good idea.

My preference goes towards trying to get (1) right along this PR so that we can ship a properly functioning feature; however, if that turns out to be a significant effort, I'm fine with outsourcing it into a follow-up PR as well. (2) and (3) should be handled by separate issues and PRs IMHO.

sybrandy:
@gambol99 The new change looks good.

@timoreimann First, I'd have to double-check, but I think all of the marathon nodes can handle requests via the REST API.  We'll hit the local Marathon node using curl when we add/update tasks.  (We store the config in JSON config files so we can easily move them between systems and keep the changes in source control.)  So, load balancing may help prevent one node from being overwhelmed.  However, like I said before, I don't know if it's an issue except in extreme circumstances.

As for separating things out into separate issues, that's a good idea.  I only mentioned it here since it was related to the changes.

gambol99:
> I also think that we should fix markDown, which could be as easy as passing the node/index to the function to make sure we don't run into the race.

Yep, changed the PR to pass the endpoint back to the [markDown()](https://github.com/gambol99/go-marathon/pull/231/commits/904b54cd8ca13c6be003302ddc1bfc3623100de1)

timoreimann:
>First, I'd have to double-check, but I think all of the marathon nodes can handle requests via the REST API.

Every Marathon node can serve API requests; all but the master just act as proxies, however. It's documented [here] (https://mesosphere.github.io/marathon/docs/core-architecture.html), and you can observe the behavior  in every follower's log file.

>As for separating things out into separate issues, that's a good idea. I only mentioned it here since it was related to the changes.

I realize I may have come across as dismissive -- sorry if that was the impression I was giving. I very much appreciate us discussing these topics and meant to say that we should give them the time they deserve in separate issues. I think we're all on the same page here. ☺️

cluster.go
```diff
@@ -27,192 +26,152 @@ import (
 )
 
 const (
-	marathonNodeUp   = 0
-	marathonNodeDown = 1
+	memberStatusUP   = 0
```
timoreimann: `s/UP/Up/` :-)

timoreimann:
@sybrandy I have to agree with you that load balancing could help us from hammering a single follower. In the end though, all traffic ends up at the leader. Since every follower must be able to act as a leader, every node must be able to sustain the complete traffic AFAICS.

sybrandy:
@timoreimann No, I didn't think you were dismissive :)

I too am glad we can bring stuff up to discuss here and give them some attention.

gambol99:
cool ... it everyone is happy, i'll merge .. @timoreimann @sybrandy ... The backoff on the node health check we can add in another PR

timoreimann:
I only glimpsed over the PR so far. Give me tonight for a more thorough look. :-)

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- [#231][PR231] Fixed Marathon cluster code
```
timoreimann: Still missing a link at the bottom of the page like:

`[PR231]: https://github.com/gambol99/go-marathon/pull/231`

cluster.go
```diff
@@ -27,192 +26,152 @@ import (
...
 }
 
-// String returns a string representation of the cluster
-func (r *marathonCluster) String() string {
-	return fmt.Sprintf("url: %s|%s, members: %s, size: %d, active: %s",
-		r.protocol, r.url, r.members, r.size, r.active)
-}
-
-type marathonNode struct {
+// member represents a individual endpoint
```
timoreimann: grammar: s/a/an/

cluster.go
```diff
@@ -27,192 +26,152 @@ import (
...
+	// step: extract and basic validate the endpoints
+	var members []*member
+	var defaultProto string
 
-	return fmt.Sprintf("member: %s:%s", member.hostname, status)
-}
+	for _, endpoint := range strings.Split(marathonURL, ",") {
+		// step: check for nothing
+		if endpoint == "" {
+			return nil, ErrInvalidEndpoint
```
timoreimann: We could improve the user experience here by turning `ErrInvalidEndpoint` into a struct that takes an error description. That way, users would know what kind of error they produced.

Could also happen in a separate PR. Leave that up to you.

cluster.go
```diff
@@ -27,192 +26,152 @@ import (
...
+	// step: wait for the node to become active ... we are assuming a /ping is enough here
+	for {
+		res, err := c.client.Get(fmt.Sprintf("%s/ping", node.endpoint))
+		if err == nil && res.StatusCode == 200 {
+			break
+		}
+		<-time.After(time.Duration(5 * time.Second))
+	}
+	// step: mark the node as active again
+	c.Lock()
```
timoreimann: We could also pull this section right into the 200 block (i.e., line 130) and return from the function afterwards.

cluster_test.go
```diff
@@ -24,54 +24,107 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestUrl(t *testing.T) {
-	cluster, _ := newCluster(http.DefaultClient, fakeMarathonURL)
-	assert.Equal(t, cluster.URL(), fakeMarathonURL)
-}
-
 func TestSize(t *testing.T) {
 	cluster, _ := newCluster(http.DefaultClient, fakeMarathonURL)
```
timoreimann: Not sure why we are ignoring the error return value here and in many other related tests. I'd probably just `assert.NotNil` on the value.

cluster_test.go
```diff
@@ -24,54 +24,107 @@ import (
...
-	assert.Equal(t, len(cluster.Active()), 3)
-	cluster.MarkDown()
-	cluster.MarkDown()
-	assert.Equal(t, len(cluster.Active()), 1)
+	assert.Equal(t, len(cluster.activeMembers()), 3)
+
+	members := cluster.activeMembers()
+	cluster.markDown(members[0])
+	cluster.markDown(members[1])
+	assert.Equal(t, 1, len(cluster.activeMembers()))
```
timoreimann: This check and `healthCheckNode` seem to race here. Not exactly sure how to solve this easily though.

cluster_test.go
```diff
@@ -24,54 +24,107 @@ import (
...
+			Expect: []string{"http://127.0.0.1:8080", "https://127.0.0.2"},
+		},
+		{
+			URL:    "http://127.0.0.1:8080/path1,127.0.0.2/path2",
+			Expect: []string{"http://127.0.0.1:8080/path1", "http://127.0.0.2/path2"},
+		},
+	}
+	for i, x := range cs {
+		c, err := newCluster(http.DefaultClient, x.URL)
+		if err != nil {
```
timoreimann: You could also write

```go
if !assert.NoError(t, err, "case %d should not have thrown and error: %s") {
  continue
}
```

cluster_test.go
```diff
@@ -24,54 +24,107 @@ import (
...
+			t.Errorf("case %d should not have thrown and error: %s\n", i, err)
+			continue
+		}
+		assert.Equal(t, x.Expect, c.activeMembers(), "case %d, expected: %v, got: %s", i, x.Expect, c.activeMembers())
+	}
+}
+
+func TestValidClusterHosts(t *testing.T) {
```
timoreimann: Aren't we testing this (mostly) in `TestClusterHosts` already?

cluster_test.go
```diff
@@ -83,6 +136,8 @@ func TestInvalidHosts(t *testing.T) {
 		"foo://127.0.0.1:3000",
 	} {
 		_, err := newCluster(http.DefaultClient, invalidHost)
-		assert.Equal(t, err, ErrInvalidEndpoint, "undetected invalid host: %s", invalidHost)
+		if err == nil {
```
timoreimann: `assert.Error()` would also work.

(Note: I'm not a huge fan of `stretchr/testify` but think we should be consistent with the existing assertion style where possible.)

CHANGELOG.md
gambol99: added to the bottom

cluster.go
gambol99: hmmm ... i wasn't too keen on placing a defer into a loop and my editor was throwing a vetting error ... Technically, its fine as I'd be doing a return straight after ...

cluster_test.go
gambol99: indeed! ... I wasn't sure how to fix it yeah ...

error.go
```diff
@@ -44,6 +44,19 @@ const (
 	ErrCodeUnknown
 )
 
+// ErrInvalidEndpoint indicates a endpoint error in the marathon urls
+type ErrInvalidEndpoint struct {
```
timoreimann: Conventionally, error types end in `Error`, so this one should be called `InvalidEndpointError` ([reference](https://github.com/golang/go/wiki/Errors#naming)).

Would make `newInvalidEndpointError` fit more consistently as well.

cluster.go
timoreimann: Interestingly -- Atom doesn't give me any vetting error.

Anyways, it doesn't matter too much. Happy to keep things the way they are now.

gambol99:
changed ErrInvalidEndpoint -> InvalidEndpointError to keep inline with naming conventions ...

timoreimann:
LGTM! Let's squash-merge this one. :-)

PR #239, opened 2016-11-09
Fix scheme-less endpoint with port.

### LINKED ISSUE #238: multiple endpoints now require protocol specifier on every host
PR #231 (supposedly inadvertently) introduced a change to the way the protocol specifier needs to be given when using multiple endpoints: Prior to that PR, one would use

```http://host1:8080,host2:8080```

which gives me

`Failed to create a client for marathon, error: endpoint: [host2:8080] must have a schema and host`

Now, we need

```http://host1:8080,http://host2:8080```

I _could_ imagine that one uses multiple hosts with varying protocols (e.g., HTTP vs. HTTPS). Since the change came in unexpectedly and breaks clients, however, I think we should return to the old format for the time being.

timoreimann:
A Marathon URL like `http://host1,host2:8080` would not be accepted because `host:8080` is considered as an opaque URL by url.Parse with the scheme being `host` and the opaque part being `8080`. We fix this by prepending the default scheme if URL.Opaque is non-empty.

Other drive-by changes:

- Panic on a parse error when we add the default scheme. (This should   not happen, hence the panic to indicate an invariant.)
- Remove a check for empty scheme. (We should have ruled out this possibility.)

Fixes #238.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- [#239][PR239] Fix scheme-less endpoint with port.
 
 ## [0.5.0] - 2016-11-07
 ### Fixed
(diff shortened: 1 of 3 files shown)


timoreimann:
@gambol99, @xperimental: Please review.

gambol99:
eeekkk ... my bad!!, I thought I catered for this ... LGTM

timoreimann:
No worries -- this one was a bit tricky to spot during development.

Merging...

PR #244, opened 2016-11-16
add taskStats support

matt-deboer:
Adds support for `?embed=apps.taskStats` optional parameter


application.go
```diff
@@ -56,41 +56,42 @@ type Port struct {
...
+	BackoffSeconds        *float64               `json:"backoffSeconds,omitempty"`
+	BackoffFactor         *float64               `json:"backoffFactor,omitempty"`
+	MaxLaunchDelaySeconds *float64               `json:"maxLaunchDelaySeconds,omitempty"`
+	Deployments           []map[string]string    `json:"deployments,omitempty"`
+	Dependencies          []string               `json:"dependencies"`
+	TasksRunning          int                    `json:"tasksRunning,omitempty"`
+	TasksStaged           int                    `json:"tasksStaged,omitempty"`
+	TasksHealthy          int                    `json:"tasksHealthy,omitempty"`
+	TasksUnhealthy        int                    `json:"tasksUnhealthy,omitempty"`
+	TaskStats             *map[string]*TaskStats `json:"taskStats,omitempty"`
```
timoreimann: I think we should be able to remove all pointers as this is a read-only parameter.

application.go
```diff
@@ -130,6 +131,17 @@ type DeleteAppOpts struct {
 	Force bool `url:"force,omitempty"`
 }
 
+// TaskStats is a container for Stats
+type TaskStats struct {
+	Stats *Stats `json:"stats,omitempty"`
+}
+
+// Stats is a collection of aggregate statistics about an application's tasks
+type Stats struct {
```
timoreimann: Depointerize both struct fields here too.

tests/rest-api/methods.yml
```diff
@@ -203,6 +203,227 @@
         }
     ]
     }
+- uri: /v2/apps?embed=apps.taskStats
+  method: GET
+  content: |
+    {
+    "apps": [
+        {
+            "args": null,
```
timoreimann: Could you please reduce the test data definition to just the fields we actually need for testing?

matt-deboer: tried to reduce further than this, but general app deserialization fails at that point.

matt-deboer:
ok; done...look good?

tests/rest-api/methods.yml
```diff
@@ -203,6 +203,102 @@
...
+            "mem": 64.0,
+            "ports": [],
+            "requirePorts": false,
+            "storeUrls": [],
+            "tasksRunning": 2,
+            "tasksStaged": 0,
+            "taskStats": {
+                "startedAfterLastScaling": {
+                    "stats": {
+                    "counts": {
```
timoreimann: This block needs to be indented one more level.

tests/rest-api/methods.yml
```diff
@@ -203,6 +203,102 @@
...
+                    },
+                    "lifeTime": {
+                        "averageSeconds": 17024.575,
+                        "medianSeconds": 17024.575
+                    }
+                    }
+                },
+                "withLatestConfig": {
+                    "stats": {
+                    "counts": {
```
timoreimann: Indent here too.

tests/rest-api/methods.yml
```diff
@@ -203,6 +203,102 @@
...
+                    },
+                    "lifeTime": {
+                        "averageSeconds": 17024.575,
+                        "medianSeconds": 17024.575
+                    }
+                    }
+                },
+                "totalSummary": {
+                    "stats": {
+                    "counts": {
```
timoreimann: Indent.

tests/rest-api/methods.yml
```diff
@@ -203,6 +203,102 @@
...
+                },
+                "withLatestConfig": {
+                    "stats": {
+                    "counts": {
+                        "staged": 0,
+                        "running": 1,
+                        "healthy": 1,
+                        "unhealthy": 0
+                    },
+                    "lifeTime": {
```
timoreimann: Indent here too.

tests/rest-api/methods.yml
```diff
@@ -203,6 +203,102 @@
...
+                },
+                "totalSummary": {
+                    "stats": {
+                    "counts": {
+                        "staged": 0,
+                        "running": 1,
+                        "healthy": 1,
+                        "unhealthy": 0
+                    },
+                    "lifeTime": {
```
timoreimann: Indent here too.

tests/rest-api/methods.yml
timoreimann: That should only be the case when you are producing incorrect JSON. (Which can easily happen when you modify `methods.yml`; at least, it keeps happening to me.)

I could successfully run your tests with the following, minimalistic JSON response:

```json
{
    "apps": [
        {
            "taskStats": {
                "startedAfterLastScaling": {
                    "stats": {
                      "counts": {
                          "staged": 0,
                          "running": 1,
                          "healthy": 1,
                          "unhealthy": 0
                      },
                      "lifeTime": {
                          "averageSeconds": 17024.575,
                          "medianSeconds": 17024.575
                      }
                    }
                }
            }
        }
    ]
    }
```

application.go
```diff
@@ -130,6 +131,17 @@ type DeleteAppOpts struct {
...
 
+// TaskStats is a container for Stats
+type TaskStats struct {
+	Stats Stats `json:"stats"`
+}
+
+// Stats is a collection of aggregate statistics about an application's tasks
+type Stats struct {
+	Counts   map[string]int     `json:"counts"`
+	LifeTime map[string]float64 `json:"lifetime"`
```
timoreimann: I'm not a native English speaker but shouldn't this be written `Lifetime`?

tests/rest-api/methods.yml
timoreimann: That should only be the case when you are producing incorrect JSON. (Which can easily happen when you modify `methods.yml`; at least, it keeps happening to me.) JSON deserialization only cares about JSON structure, missing fields should just be skipped.

I could successfully run your tests with the following, minimalistic JSON response:

```json
{
    "apps": [
        {
            "taskStats": {
                "startedAfterLastScaling": {
                    "stats": {
                      "counts": {
                          "staged": 0,
                          "running": 1,
                          "healthy": 1,
                          "unhealthy": 0
                      },
                      "lifeTime": {
                          "averageSeconds": 17024.575,
                          "medianSeconds": 17024.575
                      }
                    }
                }
            }
        }
    ]
    }
```

matt-deboer: you're right; thanks. not sure why that was failing for me ealier...

application.go
matt-deboer: yes, I would normally write it that way; though it seems that `lifeTime` is the json produced by the marathon API...happy to change the go field name though

timoreimann: If the JSON field name uses a capital T, then I agree we should have it in the variable name as well. You did not use a capital T in the JSON tag name though -- is that a typo?

Whatever the right spelling is, we should keep it in sync.

application_test.go
```diff
@@ -297,6 +297,20 @@ func TestApplications(t *testing.T) {
...
+
+	v := url.Values{}
+	v.Set("embed", "apps.taskStats")
+	applications, err := endpoint.Client.Applications(v)
+	assert.NoError(t, err)
+	assert.NotNil(t, applications)
+	assert.Equal(t, len(applications.Apps), 1)
+	assert.NotNil(t, applications.Apps[0].TaskStats)
+	assert.Equal(t, applications.Apps[0].TaskStats["startedAfterLastScaling"].Stats.Counts["healthy"], 1)
+}
```
timoreimann: Seems like this comment got lost somehow:

Can we add an assertion for TaskStats.LifeTime too please?

timoreimann:
LGTM -- thanks for bearing with me. :-)

Issue #171, opened 2016-07-18
Marathon application schema changes.


dmajere:
`taskKillGracePeriodSeconds` field was added to marathon application struct.
Documentation about it can be found here - https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_create.md

Also `secrets` field was added, but i can't find any documentation about it right now, its only mentioned in schema - https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json

timoreimann:
Both fields are related to PRs assigned to milestone 1.3. Since Mesosphere has decided to skip 1.2, I think we can expect the next stable release to include proper documentation.

I'd rather wait to add support for either in go-marathon until the features truly end up in the next stable release. Happy to review any PRs of yours though once things have landed.

timoreimann:
Support for the task kill grace period has been added by now.

timoreimann:
#299 brought support for secrets. This issue is resolved.

PR #246, opened 2016-11-18
+ add taskKillGracePeriodSeconds

kulikov:
Available since marathon 1.3.0

Docs https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/app.raml#L237

ref to #171


gambol99:
Cheers for the PR @kulikov  ... If this is a configurable option, perhaps we can add getters and setters? ... similar to https://github.com/gambol99/go-marathon/blob/master/application.go#L164-L259 ...

kulikov:
done

kulikov:
Only 3 methods has `Set` prefix: SetApplicationVersion, SetDiscovery and SetIPAddressPerTask.

But yes, we can ^)

timoreimann:
I should have said `Set` or `Add`. The newer methods normally have it.

kulikov:
ok, thanks

timoreimann:
Sorry, I only realize now that we have extended the JSON methods by the new field but don't use it anywhere for testing. IMHO, we should either remove them from the yaml file or add a deserialization test assertion.

kulikov:
done. at first i just add field everywhere, like maxLaunchDelaySeconds

timoreimann:
LGTM -- thanks a lot!

PR #253, opened 2016-12-28
Add support for secrets (DCOS enterprise)

matt-deboer:
Adds support for secrets (available in DCOS enterprise). Without this update, this client cannot be used with the enterprise version as it will fail to parse any Application definitions that make references to secrets.

See the [docs](https://docs.mesosphere.com/1.8/administration/secrets/use-secrets/) for details.


matt-deboer:
Fixes #252

application.go
```diff
@@ -344,11 +357,44 @@ func (r *Application) AddEnv(name, value string) *Application {
...
+		r.EmptyEnvs()
+	}
+	(*r.Env)[name] = EnvSecretValue{Secret: secret}
+
+	return r
+}
+
+// AddSecret adds a secret declaration
+//		name:	the name of the variable
+//		source:	the source id of the secret
```
timoreimann: s/id/ID/

application.go
```diff
@@ -344,11 +357,44 @@ func (r *Application) AddEnv(name, value string) *Application {
 	return r
 }
 
+// AddEnvSecret adds an environment variable secret to the application
+//		name:	the name of the variable
+//		secret:	the name of secret defined in the application's Secrets
+func (r *Application) AddEnvSecret(name, secret string) *Application {
```
timoreimann: Can we define this function after `AddSecret` to match the order of the underlying structs defined above?

application.go
```diff
@@ -344,11 +357,44 @@ func (r *Application) AddEnv(name, value string) *Application {
...
-	r.Env = &map[string]string{}
+	r.Env = &map[string]interface{}{}
+
+	return r
+}
+
+// EmptySecrets explicitly empties the secrets -- use this if you need to empty
+// the secrets of an application that already has secrets set (setting secrets to nil will
+// keep the current value)
+func (r *Application) EmptySecrets() *Application {
```
timoreimann: Please group with the corresponding add method.

application.go
```diff
@@ -56,43 +56,44 @@ type Port struct {
...
-	Fetch                      *[]Fetch             `json:"fetch,omitempty"`
-	IPAddressPerTask           *IPAddressPerTask    `json:"ipAddress,omitempty"`
+	ID                         string                  `json:"id,omitempty"`
+	Cmd                        *string                 `json:"cmd,omitempty"`
+	Args                       *[]string               `json:"args,omitempty"`
+	Constraints                *[][]string             `json:"constraints,omitempty"`
+	Container                  *Container              `json:"container,omitempty"`
+	CPUs                       float64                 `json:"cpus,omitempty"`
+	Disk                       *float64                `json:"disk,omitempty"`
+	Env                        *map[string]interface{} `json:"env,omitempty"`
```
timoreimann: I'm not happy that we are using `interface` now. Is there a way to avoid it and bring back type safety?

tests/rest-api/methods.yml
```diff
@@ -106,7 +106,12 @@
             "dependencies": [],
             "deployments": [],
             "disk": 0.0,
-            "env": {},
+            "env": {
+                "VAR": "VALUE",
+                "SECRET1": {
+                    "secret": "SECRET_VALUE"
```
timoreimann: Are we actually using the new values in any test? It doesn't seem so. We should probably extend/add a test.

tests/rest-api/methods.yml
```diff
@@ -128,6 +133,11 @@
                 10001
             ],
             "requirePorts": false,
+            "secrets": {
+                "SECRET1": {
```
timoreimann: Same question about test usage here.

timoreimann:
Hi @matt-deboer! Any chance you can address my comments so we can get this PR into a mergeable state?

Thanks ☺️

tests/rest-api/methods.yml
matt-deboer: ok; should be tested now

tests/rest-api/methods.yml
matt-deboer: tested now

application.go
matt-deboer: ok; PTAL

application.go
```diff
@@ -144,6 +147,47 @@ type Stats struct {
 	LifeTime map[string]float64 `json:"lifeTime"`
 }
 
+// EnvValue represents either a string environment variable value, or
+// a reference to a Secret defined in the Secrets attribute of the Application
+type EnvValue struct {
+	Value  string
+	Secret string `json:"secret"`
```
timoreimann: Instead using a `string` type and subsequently jumping through a few hoops to serialize/serialize secrets properly, how about mapping `Secret` to a struct? Combined with `omitempty` tags, this would save us both the `MarshalJSON` and `UnmarshalJSON` implementations.

(I'm not sure if it makes sense to support empty environment values, but if it does, we can and should pointerize them.)

application.go
```diff
@@ -365,7 +409,40 @@ func (r *Application) AddEnv(name, value string) *Application {
...
-	r.Env = &map[string]string{}
+	r.Env = &map[string]EnvValue{}
+
+	return r
+}
+
+// AddSecret adds a secret declaration
+//		name:	the name of the variable
+//		source:	the source ID of the secret
+func (r *Application) AddSecret(name, source string) *Application {
```
timoreimann: Question: Does each secret have to go along with an environment secret, possibly sharing certain elements? From the examples in the doc, it looks like the secret value in the environment variable has to match the key in the `secrets` map.

The reason why I'm asking: If they always need to correspond, should we populate both environment variable secrets and `secrets` by this method?

JordanFaust:
Any updates on this?

timoreimann:
@matt-deboer I see you pushing commits again, happy to see. :-)

Is this PR ready for review again?

timoreimann:
Fixes #171.

matt-deboer:
This includes some changes from @SayWhat1, though I think it goes back to the originally proposed `map[string]interface{}` for env vars...please take another look and remind us what you'd like done

timoreimann:
I'd love to see go-marathon hide the fact that Mesosphere/Marathon convolutes both environment variables and secrets into the same JSON field. That is, the API of `Env` (along with its convenience methods) should not change.

We had a similar need recently when [we added support for `UnreachableStrategy`](https://github.com/gambol99/go-marathon/pull/293). There we just iterate through all known types and pick the first one we manage to unmarshal. We could slightly adopt the approach and have custom `UnmarshalJSON` implementations on both `Env` and `Secrets`, respectively, where each tries to decode into its own type and just drops failed deserialization results. The respective `MarshalJSON` implementations would always know how to encode correctly.

WDYT?

timoreimann:
Obsoleted by #299.

PR #255, opened 2017-01-07
Check formatting (gofmt) during CI.

timoreimann:
We have seen a few PRs where`gofmt` wasn't run on changed/added code. This PR extends the Makefile by a `check-format` command that fails if there are any unformatted files, and includes the command during Travis CI runs.


timoreimann:
@gambol99 could you please have a look at let me know if this makes sense to you? My makefile-fu isn't very strong, so feel free to suggest improvements wildly. :-)

Note that all but the first commit are just to demonstrate how Travis builds look like if check-format succeeds/failed. They will be dropped from the final PR prior to merging.

Makefile
```diff
@@ -44,6 +44,10 @@ format:
 	@echo "--> Running go fmt"
 	@go fmt $(PACKAGES)
 
+check-format:
```
gambol99: yep ... or the below to show the files

```shell
check-format:
  @echo "--> Checking formatting errors"
  @gofmt -l . | grep -q '.go'; if [ $$? -eq 0 ]; then gofmt -l .; fi
```

Makefile
```diff
@@ -44,6 +44,10 @@ format:
...
+	@echo "--> Checking format"
+	@! gofmt -l . 2>&1 | read _
+
 test: deps
```
gambol99: why not make it part of the test stage? .. @$(MAKE) check-format ... so it's run every time

Makefile
timoreimann: Good idea -- done.

Makefile
timoreimann: I believe this would make it harder for users which run `make test` periodically but do not auto-gofmt their code during development since they would constantly get an error from `check-format`.

Instead, what do you think about making an `all` target like

```
all: check-format test examples
```

and invoke that from Travis CI? This way, we can offer people to just execute `make all` locally and be reassured it'll work during CI as well.

gambol99:
LGTM

PR #266, opened 2017-02-22
265: Support GPU resource

xinxian0458:
Available since Marathon 1.3.0.

### DIFF
--- a/application.go
+++ b/application.go
@@ -62,6 +62,7 @@ type Application struct {
 	Constraints                *[][]string          `json:"constraints,omitempty"`
 	Container                  *Container           `json:"container,omitempty"`
 	CPUs                       float64              `json:"cpus,omitempty"`
+	GPUs                       *float64             `json:"gpus,omitempty"`
 	Disk                       *float64             `json:"disk,omitempty"`
 	Env                        *map[string]string   `json:"env,omitempty"`
 	Executor                   *string              `json:"executor,omitempty"`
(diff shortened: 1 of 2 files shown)


gambol99:
LGTM ... @timoreimann?

application.go
```diff
@@ -62,6 +62,7 @@ type Application struct {
 	Constraints                *[][]string          `json:"constraints,omitempty"`
 	Container                  *Container           `json:"container,omitempty"`
 	CPUs                       float64              `json:"cpus,omitempty"`
+	GPUs                       float64              `json:"gpus,omitempty"`
```
timoreimann: According to the documentation, the default value for the `gpu` property is 0.

If we want users be able to reset the number of GPUs to zero, the field must be a pointer (`*float64`).

application.go
```diff
@@ -180,6 +181,13 @@ func (r *Application) CPU(cpu float64) *Application {
 	return r
 }
 
+// GPU set the amount of GPU per instance which is assigned to the application
+//		gpu:	the GPU (check MESOS docs) per instance
+func (r *Application) GPU(gpu float64) *Application {
```
timoreimann: Can we call this `GPUs` for consistency reasons?

(I know we have a `CPU()` method, but I think it should be renamed to `CPUs()` at some point too.)

application.go
```diff
@@ -180,6 +181,13 @@ func (r *Application) CPU(cpu float64) *Application {
...
+	r.GPUs = gpu
+	return r
+}
+
```
timoreimann: Could you please also add an `EmptyGPUs` method like this:

```go
func (r *Application) EmptyGPUs() *Application {
	g := 0.0
	r.GPUs = &g
	return r
}
```

This will allow users to reset the GPUs parameter to 0.

See similar functions for how to include this in your test.

application.go
xinxian0458: This change will break the build, returned message is `type Application has both field and method named GPUs`

timoreimann: Let's go with `SetGPUs` then.

xinxian0458:
done

xinxian0458:
@timoreimann Is it ok for now?

timoreimann:
@xinxian0458 looking great now. 😄 

Thanks for the contribution! 🎉

PR #267, opened 2017-02-23
Add DCOS path parameter for additional marathon instances

mariomarin:
You can install multiple instances of Marathon on DCOS [1] and you can also customize the DCOS path [2]. I'd like to be able to target these marathon instances but the path is hard-coded to **/marathon** [here](https://github.com/gambol99/go-marathon/blob/b307c574d90c358369555b8c694ac85cff4e76ac/client.go#L260):

```
		if r.config.DCOSToken != "" {
			url = fmt.Sprintf("%s/%s", member+"/marathon", uri)
		}
```

We could include a configuration parameter **DCOSPath** and change code that creates the URL to:

```
                if r.config.DCOSToken != "" && r.config.DCOSPath != "" {
                        url = fmt.Sprintf("%s/%s/%s", member, r.config.DCOSPath, uri)
                } else if r.config.DCOSToken != "" {
                        url = fmt.Sprintf("%s/%s", member+"/marathon", uri)
                }
```

The DCOS URL follows this format: *&lt;dcos-url>/service/&lt;name-of-additional-marathon-instance>*, so if your marathon instance is named *marathon-alice* you need to set the **DCOSPath** to *service/marathon-alice*.

[1] [Install Multiple Additional Marathon Instances](https://docs.mesosphere.com/1.8/usage/service-guides/marathon/install/#scrollNav-2)
[2] [Default InstallationAdminister Additional Marathon Instances via the DC/OS CLI](https://docs.mesosphere.com/1.8/usage/service-guides/marathon/install/#scrollNav-3)

### DIFF
--- a/client.go
+++ b/client.go
@@ -256,7 +256,9 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 
 		// step: Create the endpoint url
 		url := fmt.Sprintf("%s/%s", member, uri)
-		if r.config.DCOSToken != "" {
+		if r.config.DCOSToken != "" && r.config.DCOSPath != "" {
+			url = fmt.Sprintf("%s/%s/%s", member, r.config.DCOSPath, uri)
+		} else if r.config.DCOSToken != "" {
 			url = fmt.Sprintf("%s/%s", member+"/marathon", uri)
 		}
 
--- a/client.go
+++ b/client.go
@@ -197,8 +197,13 @@ func NewClient(config Config) (Marathon, error) {
 		config.PollingWaitTime = defaultPollingWaitTime
 	}
 
+	// step: if DCOSToken is set and DCOSPath is not, set DCOSPath to default value
+	if config.DCOSToken != "" && config.DCOSPath == "" {
+		config.DCOSPath = defaultDCOSPath
+	}
+
 	// step: create a new cluster
-	hosts, err := newCluster(config.HTTPClient, config.URL)
+	hosts, err := newCluster(config.HTTPClient, config.URL, config.DCOSPath)
 	if err != nil {
 		return nil, err
 	}
@@ -256,11 +261,6 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 
 		// step: Create the endpoint url
 		url := fmt.Sprintf("%s/%s", member, uri)
-		if r.config.DCOSToken != "" && r.config.DCOSPath != "" {
-			url = fmt.Sprintf("%s/%s/%s", member, r.config.DCOSPath, uri)
-		} else if r.config.DCOSToken != "" {
-			url = fmt.Sprintf("%s/%s", member+"/marathon", uri)
-		}
 
 		// step: marshall the request to json
 		var requestBody []byte
--- a/client.go
+++ b/client.go
@@ -197,13 +197,8 @@ func NewClient(config Config) (Marathon, error) {
 		config.PollingWaitTime = defaultPollingWaitTime
 	}
 
-	// step: if DCOSToken is set and DCOSPath is not, set DCOSPath to default value
-	if config.DCOSToken != "" && config.DCOSPath == "" {
-		config.DCOSPath = defaultDCOSPath
-	}
-
 	// step: create a new cluster
-	hosts, err := newCluster(config.HTTPClient, config.URL, config.DCOSPath)
+	hosts, err := newCluster(config.HTTPClient, config.URL, config.DCOSToken != "")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


client.go
```diff
@@ -256,7 +256,9 @@ func (r *marathonClient) apiCall(method, uri string, body, result interface{}) e
 
 		// step: Create the endpoint url
 		url := fmt.Sprintf("%s/%s", member, uri)
-		if r.config.DCOSToken != "" {
+		if r.config.DCOSToken != "" && r.config.DCOSPath != "" {
```
timoreimann: Instead of special-casing here, how about using `NewClient(config)` to set `r.config.DCOSPath` to the default `/marathon` (or `marathon`) if `r.config.DCOSToken` is non-empty? I'm assuming here that `DCOSToken` must always be set in DCOS Marathon; please let me know if that's not a requirement.

This would also enable us to use DCOSPath elsewhere and always have a valid value.

timoreimann (changes requested):
Can we add a test too? Thanks!

client.go
mariomarin: Done, please review it.

mariomarin:
I also added tests.

cluster.go
```diff
@@ -51,7 +51,7 @@ type member struct {
 }
 
 // newCluster returns a new marathon cluster
-func newCluster(client *http.Client, marathonURL string) (*cluster, error) {
+func newCluster(client *http.Client, marathonURL string, DCOSPath string) (*cluster, error) {
```
timoreimann: This looks out of place to me: We should not need to add a DCOS-specific parameter to the function responsible for creating the cluster configuration. Besides the design worries, it also adds a lot of noise to multiple call sites now and in the future.

What I think would be better (but not ideal; see below) is to keep adding the DCOS path in `apiCall()`. With the rest of your PR, it should boil down to something like

```
if r.config.DCOSToken != "" {		
   url = fmt.Sprintf("%s/%s/%s", member, r.config.DCOSPath, uri)
}
```

(This should work because we made sure `r.config.DCOSPath` is always set.)

Taking a step back and thinking about how we implemented the default `/marathon` path for DCOS in the first place, however, I am wondering if that was the best approach back then. In particular, does anything speak against requiring users to specify the path on the Marathon URL string? That is, shouldn't it work just as well if users passed `http://marathon1:8080/marathon,marathon2:8080/marathon,marathon3:8080/marathon` (or any path other than `marathon`) to `Config.URL`? I think this would remove a lot of extra code/boilerplate and save us from extending the `Config` struct at the price of just a little redundancy at the user's end.
The only part making things slightly more complicated is that we already default to `/marathon` if a DCOS token is given. I think we can continue to handle this properly though by checking during `apiCall()` if a DCOS token is given **and** a path is missing; in this case, we'd continue to add `/marathon` (and otherwise expect the user to have provided a custom DCOS path).

@mariomarin what do you think?

timoreimann (changes requested):
Looking at the latest PR commit, I've been thinking about whether we are taking the right direction conceptually.

Please have a look at my comment and let me know what you think.

cluster.go
mariomarin: `apiCall()` is not the only place where you need to set the DCOSPath, you also need it in a dozen of places where you make calls to the Marathon API like [this](https://github.com/gambol99/go-marathon/blob/b307c574d90c358369555b8c694ac85cff4e76ac/subscription.go#L40). I modified `newCluster()` because is where all the URL manipulation is done and it would cover all the cases.

I was just trying to keep compatibility with the existing code. I also think the user should set the URL with the **DCOSPath** included. So this is how my new patch would look like:

```Go
-		if r.config.DCOSToken != "" {
-			url = fmt.Sprintf("%s/%s", member+"/marathon", uri)
-		}
```

However, it would break compatibility with the code that @jacob-koren wrote.

@timoreimann please let me know if you want me to submit a new patch to remove the `/marathon` code.

timoreimann: The location you referenced calls `apiGet`, which in turn calls `apiCall`. All other `api*` should behave similarly.

So everything should go through `apiCall` and my previous suggestion should hopefully work. Or did I miss your point?

mariomarin: You are right, all the `api*()` behave similarly but `registerSSESubscription()` is [creating n URL](https://github.com/gambol99/go-marathon/blob/b307c574d90c358369555b8c694ac85cff4e76ac/subscription.go#L176) wich doesn't have **DCOSPath** into account.

What do you want me to do? Get rid of the Marathon path (`/marathon`) condition when **DCOSToken** is set?

timoreimann: The problem with `registerSSESubscription` (and the SSE part in general) is that it's fairly detached from the regular cluster management logic. This really needs an overhaul outside of the scope of this PR.

Nevertheless, I believe we can fix things in a minimalistic fashion by moving the cluster member selection into `buildAPIRequest` and adjust the call sites accordingly. This allows us to change `newCluster` to respect a custom DCOS path in the Marathon URL while still defaulting to `/marathon` when needed. It's unfortunate that we have to extend `newCluster`'s signature; however, I realized this function needs to change anyway along a bigger refurbishment, so I guess it's okay for now.

Hope you didn't mind that I pushed a commit of mine onto your branch. Please have a look and let me know if it makes sense to you.

mariomarin:
@timoreimann I'll test your latest commit tomorrow. Thanks for working on this.

mariomarin:
@timoreimann I tested your latest commit, and it works, please merge this pull request. Sorry for the delay, I had a couple of busy days.

timoreimann (approved):

timoreimann:
@mariomarin no worries. :) Appreciate the testing!

Thanks for your contribution and the helpful discussions!

PR #270, opened 2017-03-03
expose orignal events queue to applications

iandyh:
Hi

I found that when `go-marathon` is distributing out events to the corresponding channels, the order of the events can be different from the events out from `marathon`.

```
if event.ID&context.filter != 0 {
        context.completion.Add(1)
	        go func(ch EventsChannel, context EventsChannelContext, e *Event) {
			defer context.completion.Done()
			select {
			case ch <- e:
			case <-context.done:
				// Terminates goroutine.
			}
		}(channel, context, event)
}
```
If the channel is blocked, the execution order of the scheduling goroutine might be different from the order of the events were received. 

However, with the current design, it will be difficult to replace with a blocking call since it might block other listeners.  

I purpose the new `ListenEvents` API so that if there are applications want to preserve the order of the events they can choose to use it and filter the events at application side. 

Any feedback is much appreciated. Thanks!


timoreimann:
@iandyh thanks for the pull request. I believe we should think about ways to change the existing implementation so that events are guaranteed to be received in-order. Extending the current API surface would be a last-resort means to me.

IIUC, your approach allows for a single consumer only. Let's try to keep the level of flexibility the library offers right now while also transmitting in order.

iandyh:
@timoreimann Hi. Thanks for the feedback. The current solution is the fastest one I came up with when I found the issue. So it did lack some consideration as a client side library. 

I might have a better solution now. Let me submit it in a few days.

PR #272, opened 2017-04-07
add ignoreHttp1xx boolean field to health checks

nicgrayson:
Here is the list of fields on a health check. This should be one of them.
https://mesosphere.github.io/marathon/docs/health-checks.html


health.go
```diff
@@ -27,6 +27,7 @@ type HealthCheck struct {
 	GracePeriodSeconds     int      `json:"gracePeriodSeconds,omitempty"`
 	IntervalSeconds        int      `json:"intervalSeconds,omitempty"`
 	TimeoutSeconds         int      `json:"timeoutSeconds,omitempty"`
+	IgnoreHttp1xx          *bool    `json:"ignoreHttp1xx,ommitempty"`
```
timoreimann: Could you use all caps for HTTP in the variable name (but not the tag) please?

health.go
```diff
@@ -59,11 +60,18 @@ func (h HealthCheck) SetMaxConsecutiveFailures(i int) HealthCheck {
 	return h
 }
 
+// SetIgnoreHttp1xx sets ignore http 1xx on the health check.
+func (h HealthCheck) SetIgnoreHttp1xx(b bool) HealthCheck {
```
timoreimann: Http -> HTTP here too please.

health.go
```diff
@@ -59,11 +60,18 @@ func (h HealthCheck) SetMaxConsecutiveFailures(i int) HealthCheck {
 	return h
 }
 
+// SetIgnoreHttp1xx sets ignore http 1xx on the health check.
```
timoreimann: http -> HTTP.

health.go
```diff
@@ -59,11 +60,18 @@ func (h HealthCheck) SetMaxConsecutiveFailures(i int) HealthCheck {
...
+	h.IgnoreHttp1xx = &b
+	return h
+}
+
 // NewDefaultHealthCheck creates a default application health check
 func NewDefaultHealthCheck() *HealthCheck {
 	portIndex := 0
 	path := ""
 	maxConsecutiveFailures := 3
+	ignoreHttp1xx := false
```
timoreimann: Do we need to set it here (and use it later)? Wouldn't we just continue to apply the default by omitting the value?

health.go
nicgrayson: Yes

health.go
nicgrayson: yes

health.go
nicgrayson: Makes sense I'll fix

nicgrayson:
Fixed naming issues, amended, and force pushed.

health.go
```diff
@@ -59,6 +60,12 @@ func (h HealthCheck) SetMaxConsecutiveFailures(i int) HealthCheck {
 	return h
 }
 
+// SetIgnoreHTTP1xx sets ignore http 1xx on the health check.
+func (h HealthCheck) SetIgnoreHTTP1xx(b bool) HealthCheck {
+	h.IgnoreHttp1xx = &b
```
timoreimann: You forgot to rename this one too.

health.go
```diff
@@ -59,6 +60,12 @@ func (h HealthCheck) SetMaxConsecutiveFailures(i int) HealthCheck {
 	return h
 }
 
+// SetIgnoreHTTP1xx sets ignore http 1xx on the health check.
+func (h HealthCheck) SetIgnoreHTTP1xx(b bool) HealthCheck {
```
timoreimann: While at it: can we rename `b` to something like `ignore` too?

nicgrayson: done

health.go
nicgrayson: oops fixed

timoreimann:
LGTM, thank you Nic!

Issue #274, opened 2017-04-27
Health checks do not seem to take Basic Auth into consideration


jangie:
From my read of the code, I believe that the health checks against members in a cluster do not take into account the configured  `HTTPBasicAuthUser` and `HTTPBasicPassword` on performing the health check. 

To explain why I believe this is the case, `cluster.go`'s  `healthCheckNode` uses `http.Client.Get` (https://golang.org/src/net/http/client.go ), which does not allow for the attachment of custom headers in a similar way as `http.Client.NewRequest` and a following `http.Client.Do` would, nor does `cluster.go` seem to have a reference to the credentials at any point. `client.go` does showcase usage of the credentials as well as utilizing `http.Client.NewRequest` within `buildAPIRequest`.

If this is the case, this is problematic for consumers of this library who utilize Marathon's basic auth as basic auth seems to protect the ping endpoint in at least Marathon v1.4.2; once cluster's `markDown` is called, that member will not be able to be marked as up as it will only receive a `401` from Marathon at the `/ping` path.

timoreimann:
@jangie thanks for filing a bug report.

I believe your analysis is perfectly correct: Credentials are not taken into consideration when doing health checks against unavailable cluster nodes. To me it seems like we should create a custom struct (maybe a custom HTTP client implementation?) holding HTTP-related parameters (including the credentials) and use it whenever we try to make a request against the Marathon API.

WDYT?

jangie:
@timoreimann thank you for validating, and that sounds like a great approach. I will put together a PR to do this and will ensure it has a proper amount of test coverage. 

Thanks!

timoreimann:
Thanks @jangie, appreciated. 👍

PR #275, opened 2017-04-28
Ensure basic auth honored for healthcheck

jangie:
Related to #274 , in the case of Marathon being protected by Basic authentication, a Marathon host that is marked down would have never been marked back up once down. This specifically caused issue for my team as a dependent project (Traefik) marked a Marathon host as down, thus causing new Marathon deployments to not be monitored by that dependent project.

Explicitly removes `net/http` from `cluster`, and has it instead utilize a new struct shared by `client` that ensures normalization of request handling.

Also fixes a related potential issue if `/ping` is protected by a DCOS token.


client_test.go
```diff
@@ -39,6 +39,14 @@ func TestNewClient(t *testing.T) {
 	assert.Equal(t, conf.PollingWaitTime, defaultPollingWaitTime)
 }
 
+func TestEmptyEndpoint(t *testing.T) {
```
timoreimann: Let's call this `TestInvalidConfig` because that's what we actually care about when verifying the client.

(Specific valid and invalid URLs are covered by `cluster_test.go`.)

jangie:
@timoreimann Thanks for the comments! I've addressed them all.

I've also gone ahead and removed the existing `httpClient` property within the `client` struct, as personally I found it potentially confusing that there was both a property as well as a struct called that after addressing the comments. I have modified `subscription.go` as a result. If this is not to your liking, I can add it back and revert my change to `subscription.go`.

Thanks!

jangie:
@timoreimann Sorry about that! I should've caught that with my last go-round. Should be ready for you to take a look at again.

jangie:
I likewise don't have a DC/OS cluster to test against, and I wasn't yet able to reproduce the issue with basic credentials; we're likely taking auth off of our marathon setup shortly to reduce issues til this is merged. I can at least say that without the credentials, against a basic auth protected marathon environment, that I receive a 401 against `/ping` without credentials, so this should be a step in the right direction imo.

timoreimann:
No worries, let's just merge then. If you get a chance to test this eventually, maybe you can leave a quick note here.

Chances are we're going to bump go-marathon in Traefik soon anyway because of containous/traefik#1510; when doing so, I can piggyback this PR if you like, but also feel free to submit a dependency update yourself in Traefik if this is rather urgent for you.

jangie:
👍  thanks! i'm curious about how to bump versions of deps and so will attempt to do so just to scratch that itch.

Issue #279, opened 2017-05-08
Add support for unreachable strategy and kill selection (Marathon 1.4)


kamsz:
https://mesosphere.github.io/marathon/docs/configure-task-handling.html

kamsz:
Fixed in #280

PR #280, opened 2017-05-08
Unreachable strategy and kill selection

kamsz:
This PR adds unreachable strategy and kill selection fields (#279).


unreachable_strategy.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright 2014 Rohith All rights reserved.
```
timoreimann: Should be 2017.

kamsz:
@timoreimann done:)

Issue #212, opened 2016-10-04
Failover to other nodes in a Marathon cluster not working during event subscription


sybrandy:
So, we recently had an issue where one of our Marathon nodes was "up", but inaccessible.  Unfortunately, our code thought Marathon was down and crashed.  Looking at the logs, it looks like the AddEventListener attempted to connect to the same node several times (we use a retrier) instead of trying different nodes.  It would be better if it would attempt to connect to a different node.  A few thoughts:
1. GetMember is refactored so that it round-robins through all of the nodes regardless of whether or not the node is up or down.
2. Refactor the subscription code so that if there is an error connecting, it marks the node as down after N attempts.
3. Combine both solutions to round-robin requests between nodes and mark nodes as down if there are multiple failures when attempting to connect to them.

Right now, I have to remove the node from the config until the node comes back up.

gambol99:
hi @sybrandy  ... so just to be clear, essentially the marathon instance was responding to a /ping but ultimately not functioning? ...

timoreimann:
@gambol99 looking at the cluster code, I get the feeling that it's not doing what we expect it to do: `GetMember()` essentially [iterates through all nodes and returns the first one considered up](https://github.com/gambol99/go-marathon/blob/8908223430b37fc4106d8b8cf1f38cf7124020f1/cluster.go#L167-L170). However, we only seem to start marking nodes as down (and retrying them for liveness again) [if we have no more up nodes remaining](https://github.com/gambol99/go-marathon/blob/8908223430b37fc4106d8b8cf1f38cf7124020f1/cluster.go#L180). Looks like a chicken-and-egg problem.

I think what we should do instead is feed the result of a failing Marathon request back into the cluster, marking it as down, and trying with another host. Alternatively, we could `/ping` each host first, though that would involve an extra round trip and require some more refactoring because things like [DCOS-specific hostnames are not taken into consideration](https://github.com/gambol99/go-marathon/blob/8908223430b37fc4106d8b8cf1f38cf7124020f1/client.go#L258-L259) by `MarkDown()`.

WDYT?

sybrandy:
@gambol99 it couldn't have been responding to a /ping because, in this case, the marathon port was blocked by a firewall.  I think the other marathon nodes recognized it as being there because it was still in Zookeeper, which somehow maintained quorum.

gambol99:
[...]

PR #282, opened 2017-05-18
add SSE subscription failover

marco-jantke:
This PR tries to solve #212 

It basically works in a similar manner as the failover for common requests. It is listening to the Errors-Channel and on receiving of an error, it will mark the host as down and close the stream to the current member. Then it will try to connect to the next member in the cluster, until either a connection can be established again or no member of the Marathon cluster is up anymore.


marco-jantke:
I am still "stress-testing" this a bit and see behaviour I would like to understand before I think this PR is completely ready.

subscription_test.go
```diff
@@ -369,3 +370,59 @@ func TestEventStreamEventsReceived(t *testing.T) {
...
+	client1.hosts.members = append(client1.hosts.members, &member{endpoint: endpoint2.Server.httpSrv.URL})
+
+	events, err := endpoint1.Client.AddEventsListener(EventIDApplications)
+	require.NoError(t, err)
+
+	// This should make the SSE subscription to fail and reconnect to another cluster member
+	endpoint1.Close()
+
+	// Let a bit time so that subscription can reconnect
+	time.Sleep(10 * time.Millisecond)
```
timoreimann: As mentioned offline, would be great to have a more deterministic approach here. That is, something that tells us reliably when the subscription has been reconnected.

subscription_test.go
```diff
@@ -369,3 +370,59 @@ func TestEventStreamEventsReceived(t *testing.T) {
...
+	config := configContainer{client: &clientCfg}
+
+	endpoint := newFakeMarathonEndpoint(t, &config)
+	defer endpoint.Close()
+
+	client := endpoint.Client.(*marathonClient)
+	// Add real server as member to the cluster
+	client.hosts.members = append(client.hosts.members, &member{endpoint: endpoint.Server.httpSrv.URL})
+
+	assert.Equal(t, 2, len(client.hosts.activeMembers()), "There should be 2 active members in the cluster")
```
timoreimann: I think we should have a `require.` here, no? Otherwise, it doesn't seem meaningful to check for 1 active member in line 393.

marco-jantke: Ack, implemented.

subscription_test.go
```diff
@@ -369,3 +370,59 @@ func TestEventStreamEventsReceived(t *testing.T) {
...
+	defer endpoint2.Close()
+
+	client1 := endpoint1.Client.(*marathonClient)
+	// Add the second server to the cluster members
+	client1.hosts.members = append(client1.hosts.members, &member{endpoint: endpoint2.Server.httpSrv.URL})
+
+	events, err := endpoint1.Client.AddEventsListener(EventIDApplications)
+	require.NoError(t, err)
+
+	// This should make the SSE subscription to fail and reconnect to another cluster member
```
marco-jantke: Thanks, fixed.

marco-jantke:
@timoreimann regarding your 3rd comment, I don't see a meaningful way with the current implementation on how to accomplish this. At the moment it is completely intransparent to subscribers what happens with the underlying SSE subscription. In the test I am using the functionality like a blackbox, which I think is the right thing to do and therefore there is no way for me to detect a reconnection at the moment. AFAICS enabling this would require a larger change in the current architecture of the event subscriptions.

To make the second point of our offline discussion transparent, you asked: What happens when we run out of active Marathon cluster members? Will the SSE subscription try regularly to establish a connection again? The answer with the current state of the PR is plain: no. Basically it will encounter an error [here](https://github.com/gambol99/go-marathon/pull/282/files#diff-384bf0d04fad588e8589bdda0252018bR179) and the returned error value will run into void due to the recursive nature the subscription is retried at the moment. Note that this is different when our code path is coming from the initial subscription [here](https://github.com/gambol99/go-marathon/pull/282/files#diff-384bf0d04fad588e8589bdda0252018bR165). I will try to come up with a good solution for this. So the PR stays WIP for now.

timoreimann:
@marco-jantke 

re: "no Sleep": One possibility I see is to test `registerSSESubscription` directly as far as the failover part is concerned. That way, we can control after how many closed streams the method should stop and assert on the number of reconnects managed by an `httptest.Server`. The trade-off we'd make here is to step away from pure black box testing (i.e., go into "grey box" testing) in return for more reliable, stable tests.
I can also live with the static sleep for now, but I'd probably increase the timeout of 10 ms a bit because it's easy to miss that on a not-so-beefy machine. (Travis CI can have bad days too.)

re: "running out of Marathon nodes": I think it's okay to handle this in a separate PR.

subscription.go
```diff
@@ -188,16 +190,47 @@ func (r *marathonClient) registerSSESubscription() error {
 					r.debugLog.Printf("registerSSESubscription(): failed to handle event: %v\n", err)
 				}
 			case err := <-stream.Errors:
-				// TODO let the user handle this error instead of logging it here
-				r.debugLog.Printf("registerSSESubscription(): failed to receive event: %v\n", err)
+				r.debugLog.Printf("registerSSESubscription(): received error from SSE stream: %s", err)
+				stream.Close()
+				go r.registerSSESubscription()
```
timoreimann: Does this actually need to invoke another goroutine? Couldn't we just pull the `connectToSSE()` part into for loop starting at line 183?

subscription.go
```diff
@@ -162,19 +162,21 @@ func (r *marathonClient) registerCallbackSubscription() error {
...
 		return err
 	}
 
-	// Try to connect to stream, reusing the http client settings
-	stream, err := eventsource.SubscribeWith("", r.config.HTTPClient, request)
+	r.subscribedToSSE = true
+	return nil
+}
+
+func (r *marathonClient) registerSSESubscription() error {
```
timoreimann: How will this method behave when `RemoveEventsListener` is called? Wouldn't we need to inject some stop channel / context that we use a signal to shut down?

marco-jantke: This method was and still is not affected by `RemoveEventsListener()`. The same holds true for the http-server that is opened when you use the `EventsTransportCallback`. This is basically because the SSE connection is the underlying transport and can feed multiple `EventsListener`'s. This is of course actually a desirable behaviour of the software, but I would not consider in this PR and it is, imho, rather a nice to have. Do you think its worth to open a separate issue?

marco-jantke:
Re "no sleep": I am looking into this improvement now.
Re "running out of Marathon nodes": I fear we should tackle it in this PR. The reason is, before the failover to another node did not work, but it would always eventually reconnect to the original "master" node. At the moment we break this behaviour and I would mislike to do so. Agree?

subscription.go
timoreimann: Okay on the part that this should be handled by a different PR as I understood the method didn't "grew worse" in this regards. 👌 

AFAICS though, it's more than a nice-to-have since we seem to be leaking a goroutine per subscription. I'm thinking of applications that use go-marathon to continuously subscribe and unsubscribe to the SSE stream without ever terminating. Or did I misunderstood?

timoreimann:
@marco-jantke re: "running out of Marathon nodes": Alright, it does seem advisable to handle this case in the PR.

It seems like we "just" need one higher-level loop that continuously tries to connect to a Marathon node, establish the SSE connection, and parse events until we get an error...?

subscription.go
```diff
@@ -163,39 +163,80 @@ func (r *marathonClient) registerCallbackSubscription() error {
...
+// mustConnectToSSE will attempt to connect to the SSE stream until it can be established.
+// It panics when there is an error building the HTTP request.
+func (r *marathonClient) mustConnectToSSE() *eventsource.Stream {
+	for {
+		if len(r.hosts.activeMembers()) > 0 {
+			stream, err := r.connectToSSE()
+			if err == nil {
+				return stream
+			}
+			if err != ErrMarathonDown {
```
timoreimann: Two thoughts on this one:

1. Should we have this check at all? I'm worried about a hypothetical future case where `buildAPIRequest` (which is called by `connectToSSE`) is extended to return a non-nil error value other than `ErrMarathonDown`.
1. If we think the invariant still holds: Should it maybe rather happen in `connectToSSE`?

marco-jantke:
@timoreimann I refactored the SSE failover event subscription now once more. As you suggested in your comment I implemented now a higher-level loop that tries to reconnect continuously. Given we had an error establishing a connection to each of the active cluster members, it will pause for 5 seconds. In case of an error on the current subscription, e.g. the server closes the connection, it will immediately try to reconnect. I find this implementation much clearer. I decided for the simple 5 seconds delay to be uniform with the `healthCheckNode` function.

However it comes with the downside that also the initial connection to the cluster is happening already in a go routine and therefore connection failures are not transparent to the clients. This is the reason why I had to remove the `TestEventStreamConnectionErrorsForwarded ` test case and also that we have to wait for the connection in [this test](https://github.com/gambol99/go-marathon/pull/282/files#diff-817b4921e769ea22a692ee4ed53caab8R315) now. What do you think about the current approach?

To remove the Sleep timeout required now in the tests, we could implement e.g. a `var streamConnected chan bool` that we pass as parameter to `registerSSESubscription` and publish a bool once a connection is established. Though this would require us to go to the "grey-box" testing approach for all tests that build up a SSE connection + create a Noop receiver for the production code. What do you think is better? I am not sure whether the effort will pay out..

client.go
```diff
@@ -328,7 +338,7 @@ func (r *marathonClient) buildAPIRequest(method, uri string, reader io.Reader) (
 	// Build the HTTP request to Marathon
 	request, err = r.client.buildMarathonRequest(method, member, uri, reader)
 	if err != nil {
-		return nil, member, err
+		return nil, member, newRequestError{err}
```
marco-jantke: Not for reviewers: `buildAPIRequest` is only used on 2 places. In `apiCall` and `connectToSSE`. In order to be sure that we don't break anything with changing the error type here, I "unwrap" the error in the `apiCall` once again to keep the exact same behaviour as before. Those errors seem to be propagated until to the user. I am not 100% sure whether we really have to do this though.

subscription_test.go
```diff
@@ -17,14 +17,15 @@ limitations under the License.
...
 import (
-	"net/http"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 const eventPublishTimeout time.Duration = 250 * time.Millisecond
```
timoreimann: You can combine both consts in a single `( ... )` block.

subscription.go
```diff
@@ -162,42 +162,85 @@ func (r *marathonClient) registerCallbackSubscription() error {
...
+		httpClient := &http.Client{
+			Transport:     r.config.HTTPClient.Transport,
+			CheckRedirect: r.config.HTTPClient.CheckRedirect,
+			Jar:           r.config.HTTPClient.Jar,
+			Timeout:       r.config.HTTPClient.Timeout,
+		}
+
+		stream, err := eventsource.SubscribeWith("", httpClient, request)
+		if err != nil {
+			r.debugLog.Printf("Error subscribing to Marathon event stream: %s\n", err)
```
timoreimann: Do we need the newline at the end? FWIW, we don't use it in line 178.

Same for a few other log messages produced in this PR.

subscription_test.go
```diff
@@ -330,6 +311,9 @@ func TestEventStreamEventsReceived(t *testing.T) {
 	almostAllTestCases := testCases[:len(testCases)-1]
 	finalTestCase := testCases[len(testCases)-1]
 
+	// Let a bit time so that subscription can connect
```
timoreimann: Grammar nit-pick: "Let a bit time **pass** [...]".

subscription.go
```diff
@@ -162,42 +162,85 @@ func (r *marathonClient) registerCallbackSubscription() error {
...
+// connectToSSE tries to establish an *eventsource.Stream to any of the Marathon cluster members, marking the
+// member as down on connection failure, until there is no more active member in the cluster.
+// Given the http request can not be built, it will panic as this case should never happen.
+func (r *marathonClient) connectToSSE() (*eventsource.Stream, error) {
+	for {
+		request, member, err := r.buildAPIRequest(http.MethodGet, marathonAPIEventStream, nil)
+		if err != nil {
+			switch err.(type) {
+			case newRequestError:
+				panic(fmt.Sprintf("Unable to build new request in SSE subscription: %s\n", err.Error()))
```
timoreimann: Can we reword this more strongly towards the fact that we don't expect request build failures to happen at this stage?

Suggestion:

`"Requests for SSE subscriptions should never fail to be created: %s", err`

client.go
marco-jantke: Ok good, I removed the code to unwrap the error in the `apiCall` method again.

subscription_test.go
marco-jantke: Thanks, done.

subscription_test.go
marco-jantke: Thanks, fixed.

subscription.go
marco-jantke: I agree, this formulation is better. Implemented.

subscription.go
marco-jantke: You are right, I checked the implementation of `debugLog.Printf()` and it will always append a newline in case there is none. Removed it from logs touched by this PR.

marco-jantke:
@timoreimann thanks for the review, I implemented the points. 

As additional info: I tested this version with a real Marathon cluster, with manually shutting instances down and observing the failover mechanism. Everything worked like expected 🎉  

In case you also want to test, make sure to fix the /ping call in the `healthCheckNode`. I first need to rebase this PR in order to have the fix included. If you say its ready for that, give me your go :)

marco-jantke:
@timoreimann I had to make some adaptions to the `TestBuildApiRequestFailure` which I didn't rebase so you can have a look again. ff6aa4b

client_test.go
```diff
@@ -135,6 +135,53 @@ func TestAPIRequest(t *testing.T) {
...
+		}
+
+		_, _, err := client.buildAPIRequest("GET", test.path, nil)
+
+		if test.expectedError != nil {
+			assert.Equal(t, test.expectedError, err)
+		} else if test.expectedErrorType != nil {
+			assert.IsType(t, test.expectedErrorType, err)
+		} else {
+			assert.Fail(t, "Missing error assertion in test!")
```
timoreimann: IMHO, this should be a panic since not providing either `expectedError` nor `expectedErrorType` is a programming mistake.

We might also want to move this check to the top prior to making the call to the system-under-test, similar to how input parameter validation happens first thing in a function call.

WDYT on both points?

marco-jantke: Both valid points, thanks for the improvement suggestion. I added it as panic now to the top 👍

timoreimann:
@marco-jantke 👍 

Can you update your branch so that I can merge?

timoreimann:
Very impressive work, @marco-jantke -- thanks a lot 👏 

Merging now...

PR #284, opened 2017-05-31
[Bug] Fix build group path with force param

ge1st:
Hi @gambol99 I came across this bug for group delete and update. It was previously using the `buildPathWithForceParam` which incorrectly tacks on the `marathonAPIApps` to the url. This is a quick fix.

I've just noticed `tasks.go` uses an `addOption` utility for adding options to the url. Let me know if that's the preferred way and I can update this PR.


timoreimann:
At some point in the future, we need to improve the helper functions; but I think this is good enough for now.

Thanks for your contribution!

PR #289, opened 2017-06-14
Expose response unmarshaling error.

timoreimann:
This makes the actual error message bubble up to the caller so that it can be properly shown.


gambol99:
LGTM ...

timoreimann:
Thanks @gambol99. 🎉

PR #293, opened 2017-06-28
Support UnreachableStrategy format schema with absence reason.

timoreimann:
Undocumented as of now, but observable in [real-world installation](https://github.com/containous/traefik/issues/1743) and on the [DC/OS issue tracker](https://jira.mesosphere.com/browse/DCOS_OSS-939), the `UnreachableStrategy` field may come as a flat string-only schema alternative describing the absence reason (currently only `disabled`). We extend our struct accordingly to handle both schemas.

We also move the currently existing test into the newly created test file and replace a few `assert.`s by `require.`s.


timoreimann:
@gambol99 @marco-jantke PTAL.

@Juliens for the Traefik end. :-)

marco-jantke:
LGTM 👍

PR #295, opened 2017-07-03
Add support for containers with persistent volumes

ondrej-smola:


docker.go
```diff
@@ -46,10 +46,73 @@ type Parameters struct {
...
+	PersistentVolumeTypeRoot  PersistentVolumeType = "root"
+	PersistentVolumeTypePath  PersistentVolumeType = "path"
+	PersistentVolumeTypeMount PersistentVolumeType = "mount"
+)
+
+// PersistentVolume declares a Volume to be Persistent, and sets
+// the size (in MiB) and optional type, max size (MiB) and constraints for the Volume.
+type PersistentVolume struct {
+	Type        *PersistentVolumeType `json:"type,omitempty"`
+	Size        int                   `json:"size,omitempty"`
```
timoreimann: Does the zero value carry any semantical meaning? If yes, we need a pointer.

docker.go
```diff
@@ -98,6 +161,22 @@ func (container *Container) EmptyVolumes() *Container {
 	return container
 }
 
+// SetPersistentVolume defines persistent properties for volume
+//      size: size of the volume in MiB
+func (v *Volume) SetPersistentVolume(size int) *PersistentVolume {
```
timoreimann: Is a persistent volume with a size of zero valid to Marathon? If yes, it's appropriate to have the `size` parameter. If not, we should drop it and leave it to the user to call `SetSize` or not.

docker_test.go
```diff
@@ -122,3 +122,34 @@ func TestExternalVolume(t *testing.T) {
...
+	assert.Equal(t, 0, len(*pVol.Constraints))
+
+	container := docker.Container.Volume("/host", "/container", "RW")
+	assert.Equal(t, 1, len(*docker.Container.Volumes))
+
+	vol := (*container.Volumes)[0]
+	pVol = vol.SetPersistentVolume(256)
+	pVol.SetType(PersistentVolumeTypeMount)
+	pVol.SetMaxSize(128)
+
```
timoreimann: Please call `AddConstraint` here with multiple constraints.

docker_test.go
```diff
@@ -122,3 +122,34 @@ func TestExternalVolume(t *testing.T) {
...
+	assert.Equal(t, "/host", vol.HostPath)
```
timoreimann: Please do not validate the Volume-only parameters (i.e., drop this line and the next two). This should be the responsibility of the Volume-specific tests.

docker_test.go
```diff
@@ -122,3 +122,34 @@ func TestExternalVolume(t *testing.T) {
...
+	assert.Equal(t, "/container", vol.ContainerPath)
+	assert.Equal(t, "RW", vol.Mode)
+	assert.Equal(t, 256, vol.Persistent.Size)
+	assert.NotNil(t, vol.Persistent.Type)
```
timoreimann: Please call this like

```
if assert.NotNil(t, vol.Persistent.Type) {
  assert.Equal(t, PersistentVolumeTypeMount, *vol.Persistent.Type)
}
```

Same for line 153 (if `MaxSize` keeps being a pointer).

docker_test.go
```diff
@@ -122,3 +122,34 @@ func TestExternalVolume(t *testing.T) {
...
+	pVol.AddConstraint("bat")
+
+	assert.Equal(t, 1, len(*pVol.Constraints))
+	assert.Equal(t, []string{"bat"}, (*pVol.Constraints)[0])
+	pVol.EmptyConstraints()
+	assert.NotNil(t, pVol.Constraints)
+	assert.Equal(t, 0, len(*pVol.Constraints))
+
+	container := docker.Container.Volume("/host", "/container", "RW")
+	assert.Equal(t, 1, len(*docker.Container.Volumes))
```
timoreimann: Please use `require` instead of `assert` here.

docker_test.go
```diff
@@ -122,3 +122,34 @@ func TestExternalVolume(t *testing.T) {
 	assert.Equal(t, ev2.Name, "")
 	assert.Equal(t, ev2.Provider, "")
 }
+
+func TestDockerPersistentVolume(t *testing.T) {
+	docker := NewDockerApplication()
+
+	pVol := &PersistentVolume{Size: 2048}
```
timoreimann: I think we should drop lines 129 through 137: We want to use the convenience API calls to test our code. Thus, line 139 is how this test should start.

The constraints setup and assertions (including `EmptyConstraints`) should move below line 139 too.

docker_test.go
```diff
@@ -122,3 +122,34 @@ func TestExternalVolume(t *testing.T) {
...
+	pVol.SetMaxSize(128)
+
+	assert.Equal(t, "/host", vol.HostPath)
+	assert.Equal(t, "/container", vol.ContainerPath)
+	assert.Equal(t, "RW", vol.Mode)
+	assert.Equal(t, 256, vol.Persistent.Size)
+	assert.NotNil(t, vol.Persistent.Type)
+	assert.Equal(t, PersistentVolumeTypeMount, *vol.Persistent.Type)
+	assert.NotNil(t, vol.Persistent.MaxSize)
+	assert.Equal(t, 128, *vol.Persistent.MaxSize)
```
timoreimann: Assertions on constraints should follow after this line.

timoreimann:
Supersedes #200.

docker.go
```diff
@@ -46,10 +46,73 @@ type Parameters struct {
...
+const (
+	PersistentVolumeTypeRoot  PersistentVolumeType = "root"
+	PersistentVolumeTypePath  PersistentVolumeType = "path"
+	PersistentVolumeTypeMount PersistentVolumeType = "mount"
+)
+
+// PersistentVolume declares a Volume to be Persistent, and sets
+// the size (in MiB) and optional type, max size (MiB) and constraints for the Volume.
+type PersistentVolume struct {
+	Type        *PersistentVolumeType `json:"type,omitempty"`
```
ondrej-smola: Empty string is not a valid option, but nil is valid option an defaults to "root". We can go with requiring user to set Type or accepting current state when marathon sets it to "root" if missing. I am more inclined towards former (to be transparent about what is set) but i found latter in multiple examples [dcos letsencrypt](https://github.com/mesosphere/letsencrypt-dcos/blob/master/letsencrypt-dcos.json#L17)

docker.go
ondrej-smola: Size is required a must be positive

ondrej-smola: omitempty doe not make sense here

docker.go
```diff
@@ -46,10 +46,73 @@ type Parameters struct {
...
+	Size        int                   `json:"size,omitempty"`
+	MaxSize     *int                  `json:"maxSize,omitempty"`
```
ondrej-smola: optional and only for "mount" type, i think we can go with `int` and `omitempty`

docker.go
ondrej-smola: size must be set

docker.go
timoreimann: FWIW, the REST documentation [I cited in the original PR](https://github.com/gambol99/go-marathon/pull/200/files#r80254464) does list zero as the minimum value. Could you double-check/test this?

If the size must really be positive, we should still keep the `omitempty`. Otherwise, users will always have to set the existing value explicitly if they want to keep it.

docker.go
timoreimann: 👍

docker.go
timoreimann: Again, what if you want to update an existing persistent value but leave the size untouched? AFAICS, in this case you wouldn't want to set it.

docker_test.go
```diff
@@ -122,3 +123,32 @@ func TestExternalVolume(t *testing.T) {
...
+
+	pVol := (*container.Volumes)[0].SetPersistentVolume(256)
+	pVol.SetType(PersistentVolumeTypeMount)
+	pVol.SetMaxSize(128)
+	pVol.AddConstraint("cons1", "EQUAL", "tag1")
+	pVol.AddConstraint("cons2", "UNIQUE")
+
+	assert.Equal(t, 256, pVol.Size)
+	assert.Equal(t, PersistentVolumeTypeMount, pVol.Type)
+	assert.NotNil(t, pVol.MaxSize)
```
timoreimann: We don't need to check for nil anymore since we've converted `MaxSize`into a non-pointer.

docker.go
ondrej-smola: You are right, completely missed that point.

docker.go
timoreimann: This point seems left.

ondrej-smola: by docs size can be 0 but by my test on marathon 1.4.2 size must be positive. Moreover persistent volumes cannot be updated. This is all beta API so i would keep it to reflect latest api docs - size is required (not omitempty) but can be zero.

timoreimann: Understood, makes sense now -- thanks for the explanation. 👍

PR #297, opened 2017-07-06
Use http client with configured timeouts when none provided in config

ondrej-smola:


client.go
```diff
@@ -154,6 +155,18 @@ var (
 	ErrMarathonDown = errors.New("all the Marathon hosts are presently down")
 	// ErrTimeoutError is thrown when the operation has timed out
 	ErrTimeoutError = errors.New("the operation has timed out")
+
+	// Default http client used when creating new marathon client and no set
+	// in provided config
+	DefaultHttpClient = &http.Client{
```
gambol99: Do we need this exported?

ondrej-smola: Sry I am just used to having test package separated from main package ... we can make it private in this case.

ondrej-smola:
Thanks for catching that before merge. Will try to find solution suitable for both cases.

ondrej-smola:
By setting ResponseHeaderTimeout instead of Timeout we can set timeout only for duration between request and response headers receive. I also added check to disable sse subscription with timeout set on client.

ondrej-smola:
@timoreimann @gambol99 what do you think about it?

subscription.go
```diff
@@ -103,6 +103,12 @@ func (r *marathonClient) registerSubscription() error {
 	case EventsTransportCallback:
 		return r.registerCallbackSubscription()
 	case EventsTransportSSE:
+		if timeout := r.client.config.HTTPClient.Timeout; timeout > 0 {
```
timoreimann: Do I understand correctly that this won't allow users to ever set the timeout even if they only wanted to have it for regular (non-SSE) requests?

Is it necessary to return the error to the user? How about simply disabling the critical timeout for SSE?

subscription_test.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"net/http"
```
timoreimann: Please group the import together with the other stdlib ones above (respecting alphabetical order).

subscription.go
ondrej-smola: No, user will get error only if he configures own http client with Timeout set and starts SSE subscription. I dont think silencing timeout is good approach in this situation as it will work in most cases (as default go client) but will add undefined behaviour when connection to marathon starts to misbehave (manifested by flaky SSE test under windows) - failing fast will enable user to set appropriate timeout w.r.t. to sse implementation or create 2 clients for different purposes.

timoreimann: go-marathon users don't seem to be able to create 2 different clients right now -- AFAICS, we always reuse a possibly configured custom client for both the regular Marathon API and SSE. Thus, the only option users would have is to not set that timeout at all, which sounds like an overly strong restriction.

Or am I missing something?

timoreimann:
hey @ondrej-smola, what do you think about my latest comments, especially [this one](https://github.com/gambol99/go-marathon/pull/297#discussion_r127362822)?

subscription.go
ondrej-smola: Sorry for late reply, forgot to ack comment so it was left in pending state.

timoreimann: Ok, understood that users must not set the global timeout with this approach because it would set the read response timeout which breaks SSE. However, wouldn't users potentially want to set the read response timeout for regular (non-SSE) API requests? Or do the other timeouts they can still set somehow supersede the problematic read response timeout?

ondrej-smola: In my understanding only case that is not covered by other timeouts (set by client in this PR) is when marathon writes headers and then block sending rest of the response (or send it very slowly). We can replace client.Timeout completely by setting cancel/ctx on every non-sse request - what versions of Go are supported?

ondrej-smola: by using [context](https://blog.golang.org/context) we will also give user explicit way to cancel any pending request - this will also give user option to set per request timeouts etc. But this change is epic in relative to this. This is just an example how this can be solved in much more elegant way then just fine tunning single shared client instance.

timoreimann: We still provide support for Go 1.5. We have run into a few bugs already that are only fixed in 1.6, so we might drop support for that earliest version. I'd rather not drop support for 1.6, however.

ondrej-smola: So given Go 1.6 we can:

1. Fail when user wants to create sse subscription with client that has client.Timeout set
2. Add another http.client to Config to allow user to specify custom http.Client (config.HTTPSSEClient) that is used for sse subscriptions. This can be set to config.HTTPClient when not set. First point will also apply but then he can use config.HTTPClient with Timeout set

timoreimann: That sounds reasonable to me. 👍 

To be clear on the ordering of things that need to happen: Does that mean we'll have to deprecate 1.5 first?

ondrej-smola: AFAIK no, everything should be already in 1.5.

timoreimann: 👍

ondrej-smola:
Flaky TestMarkDown test?

gambol99:
@ondrej-smola @timoreimann ... i'll take a look at fixing the TestMarkDown ... I've restarted for the job for now

client.go
```diff
@@ -154,6 +155,24 @@ var (
 	ErrMarathonDown = errors.New("all the Marathon hosts are presently down")
 	// ErrTimeoutError is thrown when the operation has timed out
 	ErrTimeoutError = errors.New("the operation has timed out")
+
+	// Default http client used for SSE subscription requests
+	// It is invalid to set client.Timeout because it includes time to read response so
+	// set dial, tls handshake and response header timeouts instead
+	defaultHttpSSEClient = &http.Client{
```
timoreimann: The Http in the variable name should be capitalized.

client.go
```diff
@@ -154,6 +155,24 @@ var (
...
+			Dial: (&net.Dialer{
+				Timeout: 5 * time.Second,
+			}).Dial,
+			ResponseHeaderTimeout: 10 * time.Second,
+			TLSHandshakeTimeout:   5 * time.Second,
+		},
+	}
+
+	// Default http client used for non SSE requests
+	defaultHttpClient = &http.Client{
```
timoreimann: All-caps HTTP here too.

client.go
```diff
@@ -198,7 +217,16 @@ type newRequestError struct {
...
 	if config.HTTPClient == nil {
-		config.HTTPClient = http.DefaultClient
+		config.HTTPClient = defaultHttpClient
+	}
+
+	if config.HTTPSSEClient == nil {
+		config.HTTPSSEClient = defaultHttpSSEClient
+	} else if config.HTTPSSEClient.Timeout != 0 {
+		return nil, fmt.Errorf(
+			"HTTPSSEClient.Timeout cannot be set (got %v)",
```
timoreimann: I'd rephrase this slightly less technical as 

"Global timeout must not be set on custom HTTP client for SSE connections (got %s)"

client_test.go
```diff
@@ -35,10 +37,22 @@ func TestNewClient(t *testing.T) {
 
 	conf := cl.(*marathonClient).config
 
-	assert.Equal(t, conf.HTTPClient, http.DefaultClient)
+	assert.Equal(t, conf.HTTPClient, defaultHttpClient)
+	assert.Equal(t, conf.HTTPSSEClient, defaultHttpSSEClient)
```
timoreimann: Let's also assert here that the timeout on our default SSE client is not set to protect ourselves.

client_test.go
```diff
@@ -35,10 +37,22 @@ func TestNewClient(t *testing.T) {
...
 	assert.Equal(t, conf.PollingWaitTime, defaultPollingWaitTime)
 }
 
+func TestInvalidConfigHTTPSSEClientTimeoutIsSet(t *testing.T) {
+	config := Config{
+		URL: "http://marathon",
+		HTTPSSEClient: &http.Client{
+			Timeout: time.Second,
```
timoreimann: Can we say `1 * time.Second` just for clarity?

config.go
```diff
@@ -52,6 +52,8 @@ type Config struct {
 	LogOutput io.Writer
 	// HTTPClient is the http client
 	HTTPClient *http.Client
+	// HTTPClient is the http client used for sse subscriptions, can't have client.Timeout set
```
timoreimann: Name in comment does not match variable name.

timoreimann: All caps for http and sse please.

timoreimann:
@ondrej-smola code is looking good to me. 👍 

Did you have a chance to test the functionality against some real cluster? The changes are mostly integration test-ish so our unit tests won't help a lot, which is why I'm asking.

ondrej-smola:
Tested on our dev cluster (12 machines) + with dummy http server that sleeps during non sse request for long time. Also found that examples/applications seems broken with marathon 1.4.2 and 1.4.5.  It panics on https://github.com/gambol99/go-marathon/blob/master/examples/applications/main.go#L92.

timoreimann:
@ondrej-smola thanks for the note on the examples. I created issue 302 to track a fix.

timoreimann:
@gambol99 WDYT, can we merge this one?

gambol99:
LGTM ...

PR #299, opened 2017-07-08
Secrets support (DC/OS Enterprise)

SayWhat1:
Addresses comments made by @timoreimann in #253

There were enough changes that this probably merits its own PR.  Wrote custom marshal/unmarshal functions for environment variables and put them in environment_variables.go, and wrote tests in environment_variables_test.go

The Env helper function signatures didn't change and the environment variable associated with secrets gets set in its helper function.


SayWhat1:
I think maybe the EmptySecrets() should remove the secrets from Env as well as empty the secrets.  But, then would EmptyEnv() also clear the secrets, or should they be kept separate?

Let me know.

environment_variable.go
```diff
@@ -0,0 +1,56 @@
...
+
+package marathon
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// EnvironmentVariable is an environment variable associated with an application
+type EnvironmentVariable struct {
```
timoreimann: I'm wondering if we should strive for a clearer separation between regular environment variables and secrets.

AFAICS, it should be possible to maintain separate types: `map[string]string` for environment variables and `map[string]EnvSecret` for secrets. We would probably need a custom top-level unmarshaler that deserializes the `env` JSON field into a generic `map[string]interface{}`. Iterating over that, we can do type assertions to decide if a record must go into the environment variables or secrets map, respectively.

Conversely, a custom marshaler would populate a `map[string]interface{}` from each map and let encoding/json do its serialization magic.

This approach would have the added benefit that we don't even break clients that reference  `map[string]string` types directly to work with environment variables. Plus, I think it'd turn Marathon's somewhat unconventional design decision to pack different types inside the same JSON field into a cleanly separated API.

I'm currently running short on time to tweak your PR accordingly to see if the idea could work. WDYT, should we give it a try?

SayWhat1: The problem is that they aren't just overloading the environment variables, they're creating a new entry for Secrets in the app definition. See [here](https://docs.mesosphere.com/1.9/security/secrets/use-secrets/). An example config is: 

```json
{  
  "id":"/developer/service",
  "cmd":"sleep 100",
  "env":{  
     "MY_SECRET":{  
        "secret":"secret0"
     }
  },
  "secrets":{  
     "secret0":{  
        "source":"developer/my-secret"
     }
  }
}
```

The environment secret references a secret in the secrets section, which points to where the secret is stored in the secret store.  So neither the env part or the secrets part hold all the information for the secret.

I suppose the Secret could be a ```map{string}struct```.  Where the key is 'secret0' and the struct is:

```go
type struct{
    EnvVar string `json=",omitempty"`,
    Source string `json="source"`,
}
```
Then the custom unmarshaler for Env could store a secrets environment variable in the above struct, and a custom unmarshaler for Secret could assign the Source.  Then the marshaler for Env would need to be something like another struct to put the environment variable part of the secret in.

It seems a bit hacky to me, but let me know what you think.  I have some time to try to get it to work.

timoreimann:
Sorry for the long delay, life's been busy. I'll try to carve out some time over the next few days to bring this PR to a conclusion. Promised!

environment_variable.go
timoreimann: Thanks for the extra info. I've had a bit of time to ponder about the design.

From a UX perspective, my impression is that a significant number of users will not want/have to process secrets (neither from reading nor writing angle). Thus, and IMHO, it'd be convenient if we could maintain our existing environment variable map holding just non-secret environment variables and have another one holding everything related to secrets data (environment variable "pointers" and the secret sources, i.e., the kind of struct you were suggesting).

IIUC, the current design requires users interesting in processing all non-secret environment variables to change from something like

```go
for name, value := range app.Env {
    // process name/value
}
```

to 

```go
for name, kind := range app.Env {
    // Make sure this isn't a secret env var.
    if kind.Secret == "" {     
        // process name/kind.EnvVar
    }
}
```

which looks a bit clumsy and non-intuitive to me.

Personally I think it's acceptable to do a bit more heavy-lifting on the library end if the result is more easily approachable data structures to work with.

WDYT?

SayWhat1:
I don't know how I feel about how I ended up doing the marshaler/unmarshaler.  If anyone needs to do something like that again, they'll probably have to piggy back off me.  I was also looking at #300 and they're going to run into the same issue.

Please let me know what you think.

application.go
```diff
@@ -99,6 +98,8 @@ type Application struct {
 	LastTaskFailure       *LastTaskFailure        `json:"lastTaskFailure,omitempty"`
 	Fetch                 *[]Fetch                `json:"fetch,omitempty"`
 	IPAddressPerTask      *IPAddressPerTask       `json:"ipAddress,omitempty"`
+	Env                   map[string]string       `json:"env,omitempty"`
```
timoreimann: Is it intentional that we drop the pointer? This was previously required so that we could clear the environment variables. (`encoding/json` drops empty values with the `omitempty` tag.)

timoreimann: Do we actually need the `json` spec now? IIUC, we are only marshaling and unmarshaling through separate and distinct structs by now.

application.go
```diff
@@ -370,7 +378,29 @@ func (r *Application) AddEnv(name, value string) *Application {
...
+	r.Env = make(map[string]string)
+
+	return r
+}
+
+// AddSecret adds a secret declaration
+//    envvar: the name of the environment variable
+//		name:	the name of the secret
+//		source:	the source ID of the secret
+func (r *Application) AddSecret(envvar, name, source string) *Application {
```
timoreimann: Can we rename `envvar` to `envVar`?

application.go
```diff
@@ -370,7 +378,29 @@ func (r *Application) AddEnv(name, value string) *Application {
...
 // keep the current value)
 func (r *Application) EmptyEnvs() *Application {
-	r.Env = &map[string]string{}
+	r.Env = make(map[string]string)
+
+	return r
+}
+
+// AddSecret adds a secret declaration
+//    envvar: the name of the environment variable
```
timoreimann: Can we indent all three parameter descriptions on the same level?

application.go
timoreimann: Can we please return the `Env` definition to the location where it was before (i.e., line 67) in order to minimize the diff?

application.go
```diff
@@ -99,6 +98,8 @@ type Application struct {
 	LastTaskFailure       *LastTaskFailure        `json:"lastTaskFailure,omitempty"`
 	Fetch                 *[]Fetch                `json:"fetch,omitempty"`
 	IPAddressPerTask      *IPAddressPerTask       `json:"ipAddress,omitempty"`
+	Env                   map[string]string       `json:"env,omitempty"`
+	Secrets               map[string]Secret       `json:"secrets,omitempty"`
```
timoreimann: Also wondering here if we actually need a pointer to a map in order to facilitate serializing empty (i.e., cleared) secrets.

application_test.go
```diff
@@ -198,14 +198,32 @@ func TestApplicationEnvs(t *testing.T) {
...
+	app.AddEnv("hello", "world").AddEnv("foo", "bar").AddSecret("top", "secret1", "/path/to/secret")
+	assert.Equal(t, 2, len(app.Env))
+	assert.Equal(t, 1, len(app.Secrets))
+	assert.Equal(t, "world", app.Env["hello"])
+	assert.Equal(t, "bar", app.Env["foo"])
+	assert.Equal(t, "/path/to/secret", app.Secrets["secret1"].Source)
+	assert.Equal(t, "top", app.Secrets["secret1"].EnvVar)
 
 	app.EmptyEnvs()
 	assert.NotNil(t, app.Env)
```
timoreimann: If `app.Env` is a non-pointer, it can never be nil.

In case we decide to redeclare a pointer on `app.Env`: Could you please wrap the assert in an if statement and put the following statement on line 211 inside the if block? That is,

```go
if assert.NotNil(t, app.Env) {
  assert.Equal(t, 0, len(app.Env))
}
```

environment_secrets.go
```diff
@@ -0,0 +1,107 @@
...
+
+// TmpSecret holds the deserialized secrets field in a Marathon application configuration
+type TmpSecret struct {
+	Source string `json:"source,omitempty"`
+}
+
+// UnmarshalJSON unmarshals the given Application JSON into an environment variables and secrets.
+func (app *Application) UnmarshalJSON(b []byte) error {
+	aux := &struct {
+		*Alias
```
timoreimann: Does this have to a pointer?

environment_secrets.go
```diff
@@ -0,0 +1,107 @@
...
+	env := make(map[string]string)
+	secrets := make(map[string]Secret)
+
+	if aux.Env != nil {
+		for k, v := range aux.Env {
+			if s, ok := v.(string); ok {
+				env[k] = s
+				continue
+			}
+			tmp, err := json.Marshal(v)
```
timoreimann: I'm not sure I understand why you marshal here. Could you explain please?

environment_secrets.go
```diff
@@ -0,0 +1,107 @@
...
+			tmp := secrets[k]
+			tmp.Source = v.Source
+			secrets[k] = tmp
+		}
+		app.Secrets = secrets
+	}
+	return nil
+}
+
+// MarshalJSON marshals secrets into it's environment variable and secret pieces and all environment
```
timoreimann: s/it's/its/

timoreimann: Not sure I get the _and all environment_ last part. What's that supposed to mean?

environment_secrets.go
```diff
@@ -0,0 +1,107 @@
...
+	}{
+		Alias: (*Alias)(app),
+	}
+	if err := json.Unmarshal(b, aux); err != nil {
+		return err
+	}
+	env := make(map[string]string)
+	secrets := make(map[string]Secret)
+
+	if aux.Env != nil {
```
timoreimann: No need for this check: It is safe to iterate over a nil map.

environment_secrets.go
```diff
@@ -0,0 +1,107 @@
...
+			s := new(TmpEnvSecret)
+			if err := json.Unmarshal(tmp, &s); err != nil {
+				return fmt.Errorf("unrecognized environment variable type: %s", v)
+			}
+			secrets[s.Secret] = Secret{EnvVar: k}
+
+		}
+		app.Env = env
+	}
+	if aux.Secrets != nil {
```
timoreimann: No need for this check either.

environment_secrets.go
```diff
@@ -0,0 +1,107 @@
...
+	return nil
+}
+
+// MarshalJSON marshals secrets into it's environment variable and secret pieces and all environment
+// variables
+func (app *Application) MarshalJSON() ([]byte, error) {
+	env := make(map[string]interface{})
+	secrets := make(map[string]TmpSecret)
+
+	if app.Env != nil {
```
timoreimann: No need for this check.

environment_secrets.go
```diff
@@ -0,0 +1,107 @@
...
+		for k, v := range app.Env {
+			env[string(k)] = string(v)
+		}
+	}
+
+	if app.Secrets != nil {
```
timoreimann: No need for this check.

environment_secrets_test.go
```diff
@@ -0,0 +1,66 @@
...
+package marathon
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEnvironmentVariableAPI(t *testing.T) {
```
timoreimann: Isn't this test function somewhat overlapping with your extensions of `TestApplicationEnvs`? Can we eliminate the redundancy somehow?

environment_secrets_test.go
```diff
@@ -0,0 +1,66 @@
...
+	app := Application{}
+	require.Nil(t, app.Env)
+	app.AddEnv("FOO", "bar")
+	app.AddSecret("TOP", "secret", "/path/to/secret")
+	assert.Equal(t, `bar`, app.Env["FOO"])
```
timoreimann: Double quotes should suffice.

environment_secrets_test.go
```diff
@@ -0,0 +1,66 @@
...
+	assert.Equal(t, "TOP", app.Secrets["secret"].EnvVar)
+	assert.Equal(t, "/path/to/secret", app.Secrets["secret"].Source)
+
+	app.EmptyEnvs()
+	require.NotNil(t, app.Env)
+	assert.Equal(t, "", app.Env["FOO"])
```
timoreimann: I'd rather assert that the length of `app.Env` matches zero.

environment_secrets_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	app.EmptySecrets()
+	require.NotNil(t, app.Secrets)
+	assert.Equal(t, Secret{}, app.Secrets["secret"])
```
timoreimann: Length-based assertion here too.

timoreimann:
I'd really like to get another opinion on the design decisions made during this PR. Maybe @gambol99 and/or @marco-jantke would be willing to volunteer.

application.go
SayWhat1: Yes.  The way the unmarshaling is working, if they're pointers they'll throw nul dereference run time errors when ranging over them.

SayWhat1: For some reason, the json spec is needed on both.  If either are missing, it won't parse env/secrets right.

SayWhat1: Done.

application.go
SayWhat1: Same problem as ranging over env.

application.go
SayWhat1: Done.

application.go
SayWhat1: Done.

application_test.go
SayWhat1: Done.

environment_secrets.go
SayWhat1: Yes.  Otherwise NewDockerApplication() won't be inherited correctly and will cause it to return a nil app.

environment_secrets.go
SayWhat1: Removed.

environment_secrets.go
SayWhat1: Removed.

environment_secrets.go
SayWhat1: The secrets in env deserialize as a map[string]string, so instead of messing with relfection and all the error handling that requires.  I marshal it, then unmarshal it into a TmpEnvSecret.

To be honest, I don't know exactly how to use reflection to do this with possible nil values, yet, but if it makes more sense to do it that way I can work on it.

environment_secrets.go
SayWhat1: Fixed

environment_secrets.go
SayWhat1: Removed.

environment_secrets.go
SayWhat1: Removed.

environment_secrets_test.go
SayWhat1: Removed.

environment_secrets_test.go
SayWhat1: Edited.

SayWhat1:
I'm not really sure where the documentation for the secrets should go, it doesn't really fit with anything already there.

Any thoughts?

application.go
```diff
@@ -361,7 +369,7 @@ func (r *Application) AddEnv(name, value string) *Application {
 	if r.Env == nil {
 		r.EmptyEnvs()
 	}
-	(*r.Env)[name] = value
+	(r.Env)[name] = value
```
marco-jantke: The braces around `r.Env` should be unnecessary.

application.go
```diff
@@ -370,7 +378,29 @@ func (r *Application) AddEnv(name, value string) *Application {
...
+
+// AddSecret adds a secret declaration
+// envVar: the name of the environment variable
+// name:	the name of the secret
+// source:	the source ID of the secret
+func (r *Application) AddSecret(envVar, name, source string) *Application {
+	if r.Secrets == nil {
+		r.EmptySecrets()
+	}
+	(r.Secrets)[name] = Secret{EnvVar: envVar, Source: source}
```
marco-jantke: The braces around `r.Secrets` should be unnecessary.

environment_secrets.go
```diff
@@ -0,0 +1,98 @@
+/*
```
marco-jantke: Would it be clearer when the file was named something like `application_marshaling.go`? In the end it defines the marshalling / unmarshalling of a application. WDYT?

environment_secrets.go
```diff
@@ -0,0 +1,98 @@
...
+		s := new(TmpEnvSecret)
+		if err := json.Unmarshal(tmp, &s); err != nil {
+			return fmt.Errorf("unrecognized environment variable type: %s", v)
+		}
+		secrets[s.Secret] = Secret{EnvVar: k}
+	}
+	app.Env = env
+	for k, v := range aux.Secrets {
+		tmp := secrets[k]
+		tmp.Source = v.Source
```
marco-jantke: Maybe stupid point as tests seem to be fine, but how is the `Source` property filled? I only see in line 65 `Secret{EnvVar: k}` without specifying `Source: ...`. What am I missing? :D

application.go
```diff
@@ -149,6 +150,13 @@ type Stats struct {
 	LifeTime map[string]float64 `json:"lifeTime"`
 }
 
+// Secret is a reference to an existing secret object whose value may be used
```
marco-jantke: I don't understand this comment like at all :/ Can you be more precise what the "value" is here?

application.go
marco-jantke: Ah I just understood this difference is because this PR changes the `Env` to not be a pointer anymore. Is this actually right? I think this has subtle differences in JSON marshalling and also it is an API break. So do we really have to do this?

SayWhat1: It's changed back to a pointer now.

environment_secrets.go
SayWhat1: Source is set on line 70.  The EnvVar and Source of a secret come from different fields, env and secrets.

environment_secrets.go
SayWhat1: Changed it.

timoreimann:
@SayWhat1:

>I'm not really sure where the documentation for the secrets should go, it doesn't really fit with anything already there.

Hmm yeah it's tricky. I guess the least we can do is add a comment above the `Env` variable field saying something like

`# Contains non-secret environment variables. Secrets environment variables are part of the Secrets map.`

WDYT?

SayWhat1:
@timoreimann 
Sounds good to me.  Added.

application.go
timoreimann: It should work if you tell encoding/json to ignore the fields by specifying ``json:"-"``.

application_marshalling.go
```diff
@@ -0,0 +1,101 @@
...
+// TmpSecret holds the deserialized secrets field in a Marathon application configuration
+type TmpSecret struct {
+	Source string `json:"source,omitempty"`
+}
+
+// UnmarshalJSON unmarshals the given Application JSON into an environment variables and secrets.
+func (app *Application) UnmarshalJSON(b []byte) error {
+	aux := &struct {
+		*Alias
+		Env     map[string]interface{} `json:"env,omitempty"`
```
timoreimann: `omitempty` can be omitted (it's only relevant for marshaling).

application_marshalling.go
```diff
@@ -0,0 +1,101 @@
...
+		Secrets map[string]TmpSecret   `json:"secrets,omitempty"`
```
timoreimann: Here too.

application_marshalling_test.go
```diff
@@ -0,0 +1,60 @@
...
+	env := application.Env
+	secrets := application.Secrets
+
+	require.NotNil(t, env)
+	assert.Equal(t, "bar", (*env)["FOO"])
+	assert.Equal(t, "TOP", (*secrets)["secret"].EnvVar)
+	assert.Equal(t, "/path/to/secret", (*secrets)["secret"].Source)
+}
+
+func TestEnvironmentVaribleMarshal(t *testing.T) {
```
timoreimann: Typo: _Varible_

environment_secrets.go
timoreimann: We won't need reflection but a bit of type-asserting and type-casting. For the sake of comparability, I put together a working implementation and pushed it to my personal repo. Have a look at [this diff](https://github.com/SayWhat1/go-marathon/compare/secrets-support...timoreimann:299-secrets-support).

I'm on the fence on which approach I should find better (or rather less obscure): The extra marshal/unmarshal seems non-obvious (and possibly slower) while the other one is fairly low-level and far from pretty. WDYT? I can probably live with either.

environment_secrets.go
marco-jantke: I actually meant where `v.Source` is set, but now I saw it is done inside of `json.Unmarshal(b, aux)`. Thanks.

application_marshalling.go
SayWhat1: Removed.

application_marshalling.go
SayWhat1: Removed.

application_marshalling_test.go
SayWhat1: Changed.

application.go
SayWhat1: Fixed.

environment_secrets.go
SayWhat1: I have no preference either way, really.  The json marshal/unmarshal is more of a shortcut, as it uses `reflection` anyway.

Though, in the json marhsal/unmarshal, it can only be type TmpEnvVar. In your diff, we'd have to add in a check to make sure the key of envValOrSecret is equal to 'secret', but that's easy enough to add in.

I'll defer to you in this case, though.

SayWhat1:
I changed the documentation for the marshal and unmarshal functions in `application_marshalling.go` to make them more clear.  Let me know what you think.

environment_secrets.go
SayWhat1: Actually, I had some time today to do some benchmarks, and it turns out your way is about 17% faster than mine.  So, I switched it out and added a few more tests for the failure states.

application.go
```diff
@@ -149,6 +151,12 @@ type Stats struct {
 	LifeTime map[string]float64 `json:"lifeTime"`
 }
 
+// Secret is the environment variable and secret store path associated with a secret.
```
timoreimann: Maybe add another quick sentence here too saying something about `EnvVar` being populated from the regular environment variables.

application_marshalling.go
```diff
@@ -0,0 +1,109 @@
...
+	secrets := &map[string]Secret{}
+
+	for envName, genericEnvValue := range aux.Env {
+		switch envValOrSecret := genericEnvValue.(type) {
+		case string:
+			(*env)[envName] = envValOrSecret
+		case map[string]interface{}:
+			for secret, secretStore := range envValOrSecret {
+				if secret != "secret" {
+					return fmt.Errorf("expected secret but received %s", secret)
```
timoreimann: I think we should turn this into an extended condition on line 63. This way, deserialization won't break if Marathon ever extends the JSON.

application.go
SayWhat1: Changed it, let me know how you think it reads.

application_marshalling.go
SayWhat1: Done.

application_marshalling.go
```diff
@@ -0,0 +1,106 @@
...
+	Secret string `json:"secret,omitempty"`
+}
+
+// TmpSecret holds the deserialized secrets field in a Marathon application configuration
+type TmpSecret struct {
+	Source string `json:"source,omitempty"`
+}
+
+// UnmarshalJSON unmarshals the given Application JSON as expected except for environment variables and secrets.
+// Environment varialbes are stored in the Env field, and the secrets, including the environment variable part,
```
timoreimann: Typo: varialbes -> variables

timoreimann: Maybe start a new sentence after the first`field` to improve readability?

application_marshalling.go
```diff
@@ -0,0 +1,106 @@
...
+		tmp := (*secrets)[k]
+		tmp.Source = v.Source
+		(*secrets)[k] = tmp
+	}
+	app.Secrets = secrets
+	return nil
+}
+
+// MarshalJSON marshals the given Application as expected except for environment variable sand secrets,
+// which are marshaled from specalized structs.  The environment variable piece of the secrets and other
```
timoreimann: typo: specalized -> specialized

application_marshalling.go
```diff
@@ -0,0 +1,106 @@
...
+// normal environment varialbes are combined and marshaled to the env field.  The secrets and the related
```
timoreimann: typo: varialbes -> variables

application_marshalling_test.go
```diff
@@ -0,0 +1,93 @@
...
+			given:    []byte(`{"env": "value"}`),
+		},
+	}
+
+	for _, test := range tests {
+		tmpApp := new(Application)
+
+		err := json.Unmarshal(test.given, &tmpApp)
+		if assert.Error(t, err) {
+			assert.Equal(t, test.expected, err.Error()[0:23])
```
timoreimann: Can we use `strings.HasPrefix` in the assertion to be less coupled to the actual character count? Feel then free to extend the `expected` strings as you deem helpful.

This will also save us from having to include the `expected` string into the message on line 79 (right now, if that line fails, we won't be able to tell which test case was hit).

application_marshalling_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+func TestEnvironmentVariableMarshal(t *testing.T) {
+	testApp := new(Application)
+	testApp.AddSecret("TOP", "secret1", "/path/to/secret")
+	testApp.AddEnv("FOO", "bar")
+
+	tmp, err := json.MarshalIndent(testApp, "", "  ")
+	assert.Nil(t, err)
```
timoreimann: `s/assert/require/`

application_marshalling_test.go
```diff
@@ -0,0 +1,93 @@
...
+	assert.Equal(t, strings.TrimSpace(testApp.String()), string(tmp))
```
timoreimann: `testApp.String()` also calls `json.MarshalIndent`. It looks to me like we're making the same call twice and compare the results.

Shouldn't we set our expectation explicitly?

application_test.go
```diff
@@ -198,14 +198,33 @@ func TestApplicationEnvs(t *testing.T) {
 	app := NewDockerApplication()
 	assert.Nil(t, app.Env)
 
-	app.AddEnv("hello", "world").AddEnv("foo", "bar")
-	assert.Equal(t, 2, len(*app.Env))
+	app.AddEnv("hello", "world").AddEnv("foo", "bar").AddSecret("top", "secret1", "/path/to/secret")
+	assert.Equal(t, 2, len((*app.Env)))
```
timoreimann: Please wrap this in an if statement and move the assertions on the content of `app.Env` (lines 204 and 205) inside the if block.

application_test.go
```diff
@@ -198,14 +198,33 @@ func TestApplicationEnvs(t *testing.T) {
...
+	assert.Equal(t, 1, len((*app.Secrets)))
```
timoreimann: I think the secrets-related tests/assertions should only happen in `TestApplicationSecret`. WDYT?

tests/app-definitions/TestApplicationString-output.json
```diff
@@ -44,5 +40,9 @@
   "instances": 2,
   "mem": 64,
   "ports": null,
-  "dependencies": null
+  "dependencies": null,
+  "env": {
```
timoreimann: Any particular reason you moved this around? If no particular reason, then I'd prefer to have it sit at its old location in order to minimize the PR diff.

application_marshalling.go
SayWhat1: Done

SayWhat1: Done.

application_marshalling.go
SayWhat1: Done.

application_marshalling.go
SayWhat1: Done.

application_marshalling_test.go
SayWhat1: Changed it to use `strings.HasPrefix` and added a description to differentiate between the test cases.

application_marshalling_test.go
SayWhat1: Changed.

application_marshalling_test.go
SayWhat1: Yep,  Changed it to check the raw string, instead of using `testApp.String()`.

application_test.go
SayWhat1: Done.

application_test.go
SayWhat1: Yep, moved all the secret testing into `TestApplicationSecret`.

tests/app-definitions/TestApplicationString-output.json
SayWhat1: It's marshaled in the order it appears in the struct.  Since the env and secrets are last now, they need to be last in the test output or they won't test as equal.

timoreimann: 👍

application_marshalling_test.go
```diff
@@ -0,0 +1,100 @@
...
+			given:       []byte(`{"env": "value"}`),
+			description: "Bad application definition.",
+		},
+	}
+
+	for _, test := range tests {
+		tmpApp := new(Application)
+
+		err := json.Unmarshal(test.given, &tmpApp)
+		if assert.Error(t, err) {
```
SayWhat1: Yep, changed.

SayWhat1:
My pleasure.  Happy to help.

PR #301, opened 2017-07-21
Task residency support

ondrej-smola:
Described partially in [Persistent Volumes](https://mesosphere.github.io/marathon/docs/persistent-volumes.html)
[RAML type description](https://github.com/mesosphere/marathon/blob/48bfffdbb2920732ba726ce19ade6deb00121794/docs/docs/rest-api/public/api/v2/types/app.raml#L23l)

PR #305, opened 2017-08-02
residency field support is added

cihangirbesiktas:


timoreimann:
@cihangirbesiktas thanks for your PR. Did you see #301 though? It seems to accomplish the same task, and we're almost done with the review.

PR #306, opened 2017-08-03
Adds delaySeconds field to healthchecks

cihangirbesiktas:


health.go
```diff
@@ -80,6 +81,7 @@ func NewDefaultHealthCheck() *HealthCheck {
 		GracePeriodSeconds:     30,
 		IntervalSeconds:        10,
 		TimeoutSeconds:         5,
+		DelaySeconds:           15,
```
timoreimann: I'd rather not extend the default health check unless there's a really good reason to. IMHO, users should pick proper values on their own.

cihangirbesiktas: In marathon api https://mesosphere.github.io/marathon/api-console/index.html, you can find the parameter under Types->app.health.AppHealthCheck and the default value is 15. And we are using this variable in our DC/OS cluster.

timoreimann: Thanks for the pointer.

Shouldn't Marathon pick the default if `delaySeconds` isn't passed along the API request?

cihangirbesiktas: yes, but we would like to be able to change it. And we could have some app definitions that include delaySeconds and we would like to support them.

timoreimann: Sorry, I did not mean to say that we should not support the parameter (we definitely should). I'm just not sure if we should extend `NewDefaultHealthCheck()`. It looks like you could define a custom value for `DelaySeconds` by creating a health check (either an empty one or based off of `NewDefaultHealthCheck`) and set the value accordingly?

cihangirbesiktas: I think Marathon API's default value of 15 which I give as a default in NewDefaultHealthCheck() should be allright.

timoreimann:
The tests are currently failing.

health.go
marco-jantke: I would also prefer to not have it set by default. Without setting it in this client library here, we leave it to the Marathon server to change the default value without us requiring a change. AFAICS we don't loose anything through this, but gain a small little bit more resiliency against future Marathon updates. WDYT?

residency.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright 2014 Rohith All rights reserved.
```
timoreimann: Should say 2017.

cihangirbesiktas:
Removed residency since it is on another PR and made the tests pass. Could you approve and merge @timoreimann ?

timoreimann:
@cihangirbesiktas as expressed by myself and @marco-jantke, extending `NewDefaultHealthCheck` doesn't seem necessary: We apparently gain nothing nothing from it for now as we chose the current Marathon default value for `delaySeconds`, while we risk diverging from potential changes that may happen in the future.

If you think there's a good reason to keep it, I'd be grateful for an explanation. Otherwise, I'd prefer to see that part getting removed.

cihangirbesiktas:
What kind of test are you expecting?

Issue #281, opened 2017-05-11
Make sure convenience builder setters operate on pointers


timoreimann:
We often employ the builder pattern to simplify constructing objects, like this:

```
foo := application.NewFoo().AddBar().IncreaseBaz()
```

In order for this to work, each method needs to take a pointer receiver and return a pointer; otherwise, modifications will occur only on ephemeral variables that likely won't be stored anywhere.

One example where we are not doing this correctly is [HealthCheck](https://github.com/gambol99/go-marathon/blob/37fb361197f7a05f9ac0a8ca6e52ac9222ed4e4d/health.go#L34). We should find all wrong builders and fix them.

timoreimann:
Fixed by #311.

PR #311, opened 2017-08-24
Modify builder setters to operate on pointers

derek-elliott:
Addresses #281.  I'm pretty sure I found them all.

Also added tests for Health.go.


health_test.go
```diff
@@ -0,0 +1,60 @@
...
+func TestMaxConsecutiveFailures(t *testing.T) {
+	hc := new(HealthCheck)
+	hc.SetMaxConsecutiveFailures(3)
+	assert.Equal(t, 3, (*hc.MaxConsecutiveFailures))
+}
+
+func TestIgnoreHTTP1xx(t *testing.T) {
+	hc := new(HealthCheck)
+	hc.SetIgnoreHTTP1xx(true)
+	assert.Equal(t, true, (*hc.IgnoreHTTP1xx))
```
timoreimann: Could you use `assert.True` here please?

derek-elliott: Yep, done.

PR #315, opened 2017-08-25
Remove AUTHORS file.

timoreimann:
I'd like to remove the AUTHORS file for the following reasons:

- It means quite a bit of chore.
- Contributors usually do not care too much about it, so maintainers end up following up on the file themselves.
- The Makefile target only helps so much since people often use different email addresses, leading to the same author names showing up multiple times in the file.
- I don't think anybody really looks at it.
- It's a piece of information that's easy to retrieve with a tiny bit of git/shell.

@gambol99 what do you think? would you be okay with it?


gambol99:
hi @timoreimann  ... makes sense! .. github gives you all the info anyhow

PR #318, opened 2017-08-25
Default SSE HTTP client to regular client if given.

### LINKED ISSUE #309: Http client configuration no longer honored for event stream connections
I believe #297 may have introduced a non-passive change to SSE functionality. I was testing a custom build of Traefik using the master branch of go-marathon, and noticed that the SSL configuration in Traefik is no longer used when establishing the SSE connection to marathon using the latest go-marathon code.

This is because Traefik makes use of a custom HTTPClient to supply the SSL configuration, and it only knows about config.HTTPClient, not the new config.HTTPSSEClient. If a custom HTTPSSEClient is not supplied, would it make sense to default HTTPSSEClient to HTTPClient if a custom HTTPClient is supplied (instead of creating a default http sse client)?

timoreimann:
This restores compatibility for users that only pass a single, custom HTTP client.

We also move the check for an illegal global timeout to the SSE connection setup phase so that users not interested in SSE subscriptions are not bothered unnecessarily.

Fixes #309.

@marco-jantke could you review this one too please? I know you did a fair amount of work on the HTTP clients as well. Thanks!

/cc @mtweten @ondrej-smola

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,12 +59,16 @@ marathonURL := "http://10.241.1.71:8080/cluster,10.241.1.72:8080/cluster,10.241.
 
 If you specify a `DCOSToken` in the configuration file but do not pass a custom URL path, `/marathon` will be used.
 
-### Custom HTTP Clients
+### Customizing the HTTP Clients
 
-If you wish to override http clients used by the API; use cases bypassing TLS verification, load root CA's or change the timeouts etc, you can pass a custom client in the config.
-There are two clients:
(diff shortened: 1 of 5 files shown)


timoreimann:
@mtweten do you have a chance to test this branch against the latest Traefik master? I don't happen to have a TLS configuration at hand.

mtweten:
@timoreimann Will work on trying this out today!

mtweten:
@timoreimann Tested this out against Traefik master and it appears to have fixed the issue.

timoreimann:
@mtweten thanks a ton for verifying the patch. 👏 

Merging now.

PR #327, opened 2017-10-24
Marathon 1.5 support

kamsz:
Tested very briefly so far, mostly with terraform-provider-marathon.

### DIFF
--- a/application.go
+++ b/application.go
@@ -54,6 +54,11 @@ type Port struct {
 	Protocol string `json:"protocol,omitempty"`
 }
 
+// Network providers info about application networking
+type Network struct {
+	Mode string `json:"mode,omitempty"`
+}
+
 // Application is the definition for an application in marathon
 type Application struct {
 	ID          string      `json:"id,omitempty"`
@@ -64,6 +69,8 @@ type Application struct {
 	CPUs        float64     `json:"cpus,omitempty"`
 	GPUs        *float64    `json:"gpus,omitempty"`
 	Disk        *float64    `json:"disk,omitempty"`
+	Networks    *[]Network  `json:"networks,omitempty"`
+
 	// Contains non-secret environment variables. Secrets environment variables are part of the Secrets map.
 	Env                        *map[string]string  `json:"-"`
 	Executor                   *string             `json:"executor,omitempty"`
@@ -493,7 +500,7 @@ func (r *Application) CheckHTTP(path string, port, interval int) (*Application, 
 		return nil, ErrNoApplicationContainer
 	}
 	// step: get the port index
-	portIndex, err := r.Container.Docker.ServicePortIndex(port)
+	portIndex, err := r.Container.ServicePortIndex(port)
 	if err != nil {
 		return nil, err
 	}
@@ -516,7 +523,7 @@ func (r *Application) CheckTCP(port, interval int) (*Application, error) {
 		return nil, ErrNoApplicationContainer
 	}
 	// step: get the port index
-	portIndex, err := r.Container.Docker.ServicePortIndex(port)
+	portIndex, err := r.Container.ServicePortIndex(port)
 	if err != nil {
 		return nil, err
 	}
@@ -972,4 +979,34 @@ func (d *Discovery) AddPort(port Port) *Discovery {
 	ports = append(ports, port)
 	d.Ports = &ports
 	return d
+}
+
+// EmptyNetworks explicitly empties networks
+func (r *Application) EmptyNetworks() *Application {
+	r.Networks = &[]Network{}
+	return r
+}
+
+// Bridged sets the networking mode to bridged
+func (r *Application) Bridged() *Application {
+	if r.Networks == nil {
+		r.EmptyNetworks()
+	}
+	network := Network{Mode: "container/bridge"}
+	networks := *r.Networks
+	networks = append(networks, network)
+	r.Networks = &networks
+	return r
+}
+
+// Host sets the networking mode to host
+func (r *Application) Host() *Application {
+	if r.Networks == nil {
+		r.EmptyNetworks()
+	}
+	network := Network{Mode: "host"}
+	networks := *r.Networks
+	networks = append(networks, network)
+	r.Networks = &networks
+	return r
 }
--- a/application.go
+++ b/application.go
@@ -922,7 +922,7 @@ func (r *Application) EmptyNetworks() *Application {
 	return r
 }
 
-// Bridged sets the networking mode to bridged
+// BridgedNetwork sets the networking mode to bridged
 func (r *Application) BridgedNetwork(name string) *Application {
 	if r.Networks == nil {
 		r.EmptyNetworks()
@@ -934,7 +934,7 @@ func (r *Application) BridgedNetwork(name string) *Application {
 	return r
 }
 
-// Bridged sets the networking mode to container
+// ContainerNetwork sets the networking mode to container
 func (r *Application) ContainerNetwork(name string) *Application {
 	if r.Networks == nil {
 		r.EmptyNetworks()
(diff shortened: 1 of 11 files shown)


gambol99:
hi @kamsz ... apologizes for the delay, i'm gonna try look at this one day ..

kamsz:
@gambol99 no prob, I'm using it heavily with terraform-provider-marathon and so far it works without issues.

application.go
```diff
@@ -34,14 +34,6 @@ type Applications struct {
 	Apps []Application `json:"apps"`
 }
 
-// IPAddressPerTask is used by IP-per-task functionality https://mesosphere.github.io/marathon/docs/ip-per-task.html
-type IPAddressPerTask struct {
```
timoreimann: IP address per task seems to be removed. Is it because Marathon 1.5 does not support it anymore?

We should try hard not to break existing clients if somehow possible.

application.go
```diff
@@ -54,6 +46,12 @@ type Port struct {
 	Protocol string `json:"protocol,omitempty"`
 }
 
+// Network providers info about application networking
```
timoreimann: Typo: provides.

application.go
```diff
@@ -72,7 +72,6 @@ type Application struct {
 	Instances                  *int                `json:"instances,omitempty"`
 	Mem                        *float64            `json:"mem,omitempty"`
 	Tasks                      []*Task             `json:"tasks,omitempty"`
-	Ports                      []int               `json:"ports"`
```
timoreimann: Will removing this field break backwards compatibility?

application.go
```diff
@@ -92,14 +91,12 @@ type Application struct {
 	UpgradeStrategy       *UpgradeStrategy        `json:"upgradeStrategy,omitempty"`
 	UnreachableStrategy   *UnreachableStrategy    `json:"unreachableStrategy,omitempty"`
 	KillSelection         string                  `json:"killSelection,omitempty"`
-	Uris                  *[]string               `json:"uris,omitempty"`
```
timoreimann: Will removing this field break backwards compatibility?

application.go
```diff
@@ -493,7 +479,7 @@ func (r *Application) CheckHTTP(path string, port, interval int) (*Application,
 		return nil, ErrNoApplicationContainer
 	}
 	// step: get the port index
-	portIndex, err := r.Container.Docker.ServicePortIndex(port)
+	portIndex, err := r.Container.ServicePortIndex(port)
```
timoreimann: Backwards compatibility?

application.go
```diff
@@ -973,3 +890,45 @@ func (d *Discovery) AddPort(port Port) *Discovery {
...
+	}
+	network := Network{Name: name, Mode: "container"}
+	networks := *r.Networks
+	networks = append(networks, network)
+	r.Networks = &networks
+	return r
+}
+
+// HostNetwork sets the networking mode to host
+func (r *Application) HostNetwork(name string) *Application {
```
timoreimann: Instead of having three different methods to set the network mode, let's have a single one that takes a custom string type denoting the possible modes (which we can provide from a set of constants).

timoreimann (changes requested):
AFAICS, the PR breaks backwards compatibility for clients that are not on Marathon 1.5 and haven't adjusted to the API changes yet. I think we should try hard to make the new features additive without removing the old functionality right away.

WDYT?

timoreimann:
Hey there @kamsz, do you possibly have any news?

kamsz:
@timoreimann Hi, not yet. I didn't have time to work on that this week. I'll try to look into this next week.

kamsz:
@timoreimann I've tried my best to keep backward compatibility. I'll try to fix tests tomorrow. Could you please look at the changes and let me know what you think?

kamsz:
I've fixed some tests. Not sure how to fix a group test, since methods from methods.yml would collide. Any ideas? @timoreimann @gambol99

application_test.go
```diff
@@ -42,6 +42,32 @@ func TestApplicationMemory(t *testing.T) {
 	assert.Equal(t, 50.0, *app.Mem)
 }
 
+func TestNewApplicationString(t *testing.T) {
```
timoreimann: Could you please try to merge this test with `TestApplicationString` in a table-driven fashion like [we also do it here](https://github.com/gambol99/go-marathon/blob/3a8544a7c28ec83645fc96feff75bc3174e426aa/unreachable_strategy_test.go#L96-L129)?

I think the test struct should come with a `func(app *Application)` that does the version-specific modifications, and another string specifying the golden test data to compare against.

Please also try to do so for the other newly created tests that are similar enough to existing tests.

docker.go
```diff
@@ -349,6 +398,24 @@ func (docker *Docker) EmptyParameters() *Docker {
 	return docker
 }
 
+// ContainerServicePortIndex finds the service port index of the exposed port
+//		port:			the port you are looking for
+func (container *Container) ContainerServicePortIndex(port int) (int, error) {
+	if container.PortMappings == nil || len(*container.PortMappings) == 0 {
+		return 0, errors.New("The docker does not contain any port mappings to search")
```
timoreimann: `The docker` sounds wrong -- should probably be `The container`.

docker.go
```diff
@@ -349,6 +398,24 @@ func (docker *Docker) EmptyParameters() *Docker {
...
+
+	// step: iterate and find the port
+	for index, containerPort := range *container.PortMappings {
+		if containerPort.ContainerPort == port {
+			return index, nil
+		}
+	}
+
+	// step: we didn't find the port in the mappings
+	return 0, fmt.Errorf("The container port required was not found in the container port mappings")
```
timoreimann: Let's mention the container `port` we intend to find.

docker.go
```diff
@@ -255,6 +257,19 @@ func (docker *Docker) Host() *Docker {
 	return docker
 }
 
+// ExposeContainer sets the container to expose the following TCP ports
+//		ports:			the TCP ports the container is exposing
+func (container *Container) ExposeContainer(ports ...int) *Container {
```
timoreimann: I don't think we should have the `Container` part in the `ExposeContainer` method name: The method is already defined on a `Container`, so users should be able to infer the type easily (and we can't conflict).

I'm also thinking it could make it easier for people to transition from Marathon pre-v1.5 to v1.5+ if most method names stay the same. We might even reuse part of the logic. WDYT?

(Similar argumentation for other new methods also coming with a `Container` name part.)

timoreimann (changes requested):
I haven't been able to complete the review yet. Nevertheless, I wanted to leave notes on two major points that caught my attention: Using table-driven tests and dropping the `Container` part from new method names. Let me know what you think.

 I'll also look into the group testing question you had. Maybe our [Marathon fake framework](https://github.com/gambol99/go-marathon/tree/ec4a50170df729be7380d2797177dea832f41013#marathon-fake) can help with that, but I yet have to dig deeper to say for sure.

Thanks for all your work on getting go-marathon ready for 1.5! 👏  I'll get back to you as soon as I can.

kamsz:
@timoreimann I've fixed most of the comments you had. Still not sure what to do with this group test.

timoreimann (changes requested):
I left a note on the first test in the PR and would love to see my suggestions be applied there and in the other tests. Basically, I'd like us to use table-driven tests where the setup is part of each test case (i.e., defined in the fields of the associated anonymous struct).

Regarding group tests: Do I understand correctly that you'd like the tests to return a different result from a particular endpoint than what's already specified in the YAML file today? If so, then I think [scopes](https://github.com/gambol99/go-marathon/tree/ec4a50170df729be7380d2797177dea832f41013#scope) could solve your problem. Have you taken a look at that? Would it help?

I actually have a question on the group matter: In your PR I see group-related tests being amended but no production code getting added. Does that mean we have some kind of breaking change on the groups endpoint? Or did I miss something?

katallaxie:
@kamsz are you still working on that? Or should I catch up? Because I would really love to see this in Traefik 1.5

timoreimann:
@katallaxie from my perspective, I believe it's okay to carry the PR at this stage. Bringing the feature to completion would make a lot of folks quite happy. :)

PR #328, opened 2017-12-14
Simplify debug logger to single function.

timoreimann:
Even with no `ConfigOutput` set, the debug logger is fairly expensive on many calls or large input parameters due to the fact that the used log.Logger implementation requires an internal lock. We reduce contention and optimize performance by wrapping the log package with a single self-defined logging function we can configure as a no-op when needed.


timoreimann:
@marco-jantke PTAL.

timoreimann:
@marco-jantke turns out the missing lock could indeed be a problem. After talking to @gottwald, we came to the conclusion that we should go back to the `log`-based approach but continue to hide it behind a function. While the log gives us synchronization, the function enables us to turn a missing `LogOutput` parameter into a true no-op.

I think this makes most sense. PTAL again.

marco-jantke:
Great approach :+1: So I stick with my approval :)

PR #337, opened 2018-02-01
Allow force delete of deployments

fbcbarbosa:
The DeleteDeployment function was ignoring the force parameter, always making "unforced" calls.

Append the force parameter as a `?force=true` querystring to the request, and update the request to expect a nil response in this case.


deployment_test.go
```diff
@@ -83,3 +83,12 @@ func TestDeleteDeployment(t *testing.T) {
 	assert.Equal(t, id.DeploymentID, "0b1467fc-d5cd-4bbc-bac2-2805351cee1e")
 	assert.Equal(t, id.Version, "2014-08-26T08:20:26.171Z")
 }
+
+func TestDeleteDeploymentForce(t *testing.T) {
+	endpoint := newFakeMarathonEndpoint(t, nil)
+	defer endpoint.Close()
+	resp, err := endpoint.Client.DeleteDeployment(fakeDeploymentID, true)
+	assert.NoError(t, err)
+	assert.NotNil(t, t)
```
timoreimann: I don't think we need this check.

deployment_test.go
```diff
@@ -83,3 +83,12 @@ func TestDeleteDeployment(t *testing.T) {
 	assert.Equal(t, id.DeploymentID, "0b1467fc-d5cd-4bbc-bac2-2805351cee1e")
 	assert.Equal(t, id.Version, "2014-08-26T08:20:26.171Z")
 }
+
+func TestDeleteDeploymentForce(t *testing.T) {
+	endpoint := newFakeMarathonEndpoint(t, nil)
+	defer endpoint.Close()
+	resp, err := endpoint.Client.DeleteDeployment(fakeDeploymentID, true)
+	assert.NoError(t, err)
```
timoreimann: Can you please use `require.NoError` here? I don't think there's any value in checking `resp` if we have already received an error at this point.

deployment.go
```diff
@@ -108,8 +108,17 @@ func (r *marathonClient) Deployments() ([]*Deployment, error) {
...
+	if force {
+		path += "?force=true"
+		return nil, r.apiDelete(path, nil, nil)
+	}
+
 	deployment := new(DeploymentID)
-	err := r.apiDelete(fmt.Sprintf("%s/%s", marathonAPIDeployments, id), nil, deployment)
+	err := r.apiDelete(path, nil, deployment)
+
 	if err != nil {
```
timoreimann: I don't think we need the check; it should suffice return right away:

```go
return deployment, err
```

fbcbarbosa:
@timoreimann I have applied the changes you suggested to TestDeleteDeploymentForce and TestDeleteDeployment.

But about this:

```
I don't think we need the check; it should suffice to return right away:
```

This means you'd return a pointer to an empty struct in case of error. I think this could have some weird side effects. For example, if someone using the API _does not_ check for an error and try to see if their rollback worked:

```
rollback, _ := client.DeleteDeployment(dep.DeploymentID, false)  // ignore error
client.WaitOnDeployment(rollback.DeploymentID, timeout);  // success!
```

`client.WaitOnDeployment` would succeed (because it can't find the empty deployment). So the user might be under the impression the rollback worked. A panic due to a nil value would be a better behavior.

I'm not 100% up-to-date on all best practices, but it seems that upon error, the recommendation is to return a nil value: 

[Reddit discussion](https://www.reddit.com/r/golang/comments/6ffls4/return_nil_or_pointer_to_empty_struct_on_error/)
[Go stdlib example](https://golang.org/src/io/ioutil/ioutil.go?s=1464:1510#L52)

timoreimann:
@fbcbarbosa I'm usually making the assumption that once your error is not nil and you ignore it, all bets are off as far as any other parameters are concerned (unless documented otherwise).

That said, I do see your point that returning nil explicitly and thus opening the possibility for a loud and clear panic adds another "safety net" (or "concrete floor" I should say). So okay for me despite the fact that all non-error checking humans are basically lost souls. :)

PR #338, opened 2018-03-03
Handle HTTP 405 error

### LINKED ISSUE #268: DeleteApplication fails with "Marathon API Error: "
I seem to have this intermittent issue where calling DeleteApplication fails with the error `"Marathon API error: "` and no additional info. Nothing appears in the Marathon logs. 

I'm using Marathon 1.3.0 and Mesos 1.0.1.

Have you encountered this issue at all?

alenkacz:
This fixes #268

### DIFF
--- a/error_test.go
+++ b/error_test.go
@@ -61,6 +61,13 @@ func TestErrors(t *testing.T) {
 			errText:  "App '/not_existent' does not exist",
 			content:  `{"message": "App '/not_existent' does not exist"}`,
 		},
+		// 405
+		{
+			httpCode: http.StatusMethodNotAllowed,
+			errCode:  ErrCodeMethodNotAllowed,
+			errText:  "",
+			content:  `{"message": null}`,
+		},
 		// 409 POST
 		{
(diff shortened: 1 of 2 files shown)


error_test.go
```diff
@@ -61,6 +61,13 @@ func TestErrors(t *testing.T) {
 			errText:  "App '/not_existent' does not exist",
 			content:  `{"message": "App '/not_existent' does not exist"}`,
 		},
+		// 405
+		{
+			httpCode: http.StatusMethodNotAllowed,
+			errCode:  ErrCodeMethodNotAllowed,
+			errText:  "",
+			content:  `{"message": null}`,
```
timoreimann: Did you possibly verify that Marathon returns no real content? At least at the time when we created the various error codes, it usually did.

alenkacz: yep, that null thing inside json is what marathon returns when you do HTTP DELETE on /v2/apps

timoreimann: Gotcha, thanks. 👍

PR #343, opened 2018-06-06
Fixed the types in  PodUpgrade struct

riot-mraviv:
@BUGFIX Fixed the types in PodUpgrade struct, types changed from int to float64 since it should be a number between 0 and 1.


gambol99:
much appreciated @riot-mraviv

PR #352, opened 2018-11-27
Add support for kill & wipe marathon stateful task

Komey:
Add support for kill & wipe marathon stateful task, which is not support yet


gambol99:
LGTM .. thank you kindly @Komey

PR #359, opened 2019-10-24
added pullConfig

fatz:
Added pullConfig support for private registries. https://mesosphere.github.io/marathon/docs/native-docker.html#provisioning-containers-with-the-ucr


docker.go
```diff
@@ -127,6 +127,19 @@ type ExternalVolume struct {
...
+type PullConfig struct {
+	Secret string `json:"secret,omitempty"`
+}
+
+// NewPullConfig creats a *PullConfig based on a given secret
+func NewPullConfig(secret string) *PullConfig {
+	pullConfig := new(PullConfig)
+	pullConfig.Secret = secret
+
+	return pullConfig
```
timoreimann: I think we can simplify this to

```go
return &PullConfig{
  Secret: secret,
}
```

fatz: of course... will fix that

timoreimann:
Merged; thanks for your contribution. ❤️

PR #360, opened 2019-11-05
Add role to Application and Pod

br-lewis:
This adds an optional `Role` field to the `Application` and `Pod` definitions as per https://github.com/mesosphere/marathon/commit/02b775ea427d2cada6765226e2c4caf885add5d6


ArmandGrillet:
@timoreimann Could you please take a look? FYI we're starting to use go-marathon for the DC/OS CLI.

PR #362, opened 2019-11-28
Add fields to Queue Items

br-lewis:
This adds a number of missing fields to `Queue`'s `Item` struct.


br-lewis:
Looks like this also includes some commits that were already merged in to master, should I squash the PR?

PR #363, opened 2019-12-02
 Change backoff and maxlaunchdelay to floats

br-lewis:
The Marathon RAML defines both of those as `number` which means any JSON
number and so includes both floats and ints. The Marathon API will return
floats in this field which is why this needed to be changed though it
breaks the library's API
https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/pod.raml#L36-L58

### DIFF
--- a/pod_scheduling.go
+++ b/pod_scheduling.go
@@ -18,9 +18,9 @@ package marathon
 
 // PodBackoff describes the backoff for re-run attempts of a pod
 type PodBackoff struct {
-	Backoff        *int     `json:"backoff,omitempty"`
+	Backoff        *float64 `json:"backoff,omitempty"`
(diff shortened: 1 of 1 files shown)


timoreimann:
👋 

quick question for my understanding: has this always been wrong in go-marathon, or did Marathon change the type from int to number/int at some point?

br-lewis:
Near as I can tell from looking into the git history and the Marathon code, it's been a double for a while so I think go-marathon has been incorrect.
https://github.com/mesosphere/marathon/blame/bc92ff869c1c724ae1072f67ecc41734f26a08f3/src/main/scala/mesosphere/marathon/raml/PodConversion.scala#L79-L82

PR #364, opened 2019-12-20
BUG do not use omitempty on plain bool

fatz:
# The Problem
The issue occurred by creating a pod with containers that use an artifact with `extract = false`. When this is posted to marathon `omitempty` will omit the false bool from the marshalled JSON. But the default value for extract is `true` ( see https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/artifact.raml#L13 ). So reading the pod again leads to `extract = true`. With the current implementation a user never wont be able to set this to `false` as it gets omitted.

# Solution
do not use `omitempty` with bool values to ensure being able to set them to false.

### DIFF
--- a/examples/pods/main.go
+++ b/examples/pods/main.go
@@ -56,7 +56,7 @@ func createRawPod() *marathon.Pod {
 			Image: &marathon.PodContainerImage{
 				Kind:      "DOCKER",
 				ID:        "nginx",
-				ForcePull: true,
+				ForcePull: marathon.Bool(true),
 			},
 			VolumeMounts: []*marathon.PodVolumeMount{
 				&marathon.PodVolumeMount{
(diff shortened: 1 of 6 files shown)


timoreimann:
The issue with not using `omitempty` is that users then always have to set the field when the default value is not what they want. Say a user set the field in question to `true`: now, on every subsequent request, they have to keep setting it to `true` to prevent it from being reverted to `false` again.

What we did in similar cases before was to use a pointer to a bool together with `omitempty`. It ensures that the default value (`nil`) does not impact any previous non-default settings to the affected values. We also tend to provide convenience methods to set values for pointerized fields to compensate for the slight reduction in convenience.

What do you think?

fatz:
Sounds good to me... I'll update the PR

fatz:
@timoreimann I've updated the PR to use bool pointers. The caveat with bool pointers is that they are hard to maintain in code as you need to initialize a var to receive its pointer.

To improve the usability of this I've added a Bool function which returns the pointer to a bool value.

I've updated the POD example which is using `ForcePull` to use `marathon.Bool(true)` see here: https://github.com/gambol99/go-marathon/pull/364/files#diff-bd1340f691222513bc04cd892209f4edR59

PR #366, opened 2019-12-24
allow file based secrets on pods

fatz:
adding Secret for volumes to enable file based secrets support

### DIFF
--- a/volume.go
+++ b/volume.go
@@ -20,6 +20,7 @@ package marathon
 type PodVolume struct {
 	Name       string            `json:"name,omitempty"`
 	Host       string            `json:"host,omitempty"`
+	Secret     string            `json:"secret,omitempty"`
 	Persistent *PersistentVolume `json:"persistent,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


volume.go
```diff
@@ -38,6 +39,14 @@ func NewPodVolume(name, path string) *PodVolume {
 	}
 }
 
+// NewPodVolume creates a new PodVolume for file based secrets
+func NewPodVolumeSecret(name, secretPath string) *PodVolume {
```
timoreimann: Would we also need/want `PodVolume.Host` in the case of defining a pod volume for file-based secrets, or is that optional?

fatz: its not even supported. https://mesosphere.github.io/marathon/docs/secrets.html#example-configuration-of-file-based-secrets-for-pod-definitions

The flow is:
1. Define `PodVolume` to provide the secret
2. Define PodVolumeMount us use the secret and specify the container path.

Host is in this case not relevant. Furthermore AFAIK it must not be specified.

timoreimann: Gotcha, thanks for explaining.

timoreimann:
@fatz could you rebase / update the branch so that we can merge?

fatz:
@timoreimann updated and tests where running through

